PR #7, opened 2021-03-28
refactor: use the new cache endpoint

johnletey:


martonlederer:
I believe this is already implemented, so the pr can be closed @johnletey

PR #16, opened 2021-05-11
fix signature interface and wrong result

Xaber20110202:
1. add `Object.values` before pass `message.data` to `new Uint8Array`
2. change signature param interface, because of that options in signature function just passed to `crypto.subtle.sign` (https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign#parameters)
3. more specific error message


martonlederer (approved):
Checked, works great

PR #23, opened 2021-06-17
Update Drive GQL Query

vilenarios:
Fixes the graphQL query to only return public drives by including a filter for "Drive-Privacy" === "public".  This should fix the issue https://github.com/th8ta/ArConnect/issues/21


t8:
Thanks so much for the contribution here, @vilenarios!

martonlederer (approved):
LGTM! Thanks!

PR #25, opened 2021-07-06
Corrected unixTime and lastModifiedDate

vilenarios:
Corrected unixTime and lastModifiedDate for drive, root folder and file transactions to show seconds since unix epoch not milliseconds, by dividing by 1000 and rounding.

### DIFF
--- a/src/utils/archive.ts
+++ b/src/utils/archive.ts
@@ -72,7 +72,7 @@ export async function createMetadataTransaction(
       data: JSON.stringify({
         name: data.filename,
         size: getSizeBytes(data.content),
-        lastModifiedDate: data.timestamp,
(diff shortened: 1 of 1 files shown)


src/utils/archive.ts
```diff
@@ -72,7 +72,7 @@ export async function createMetadataTransaction(
       data: JSON.stringify({
         name: data.filename,
         size: getSizeBytes(data.content),
-        lastModifiedDate: data.timestamp,
+        lastModifiedDate: Math.round(data.timestamp / 1000),
```
arein, line 75: any chance we can make this an actual date? `lastModifiedDate` hints at something of type timestamp not of type `timetamp in seconds`

PR #26, opened 2021-07-08
feat: tests allowances state

arein:
Also:
* installs [jest](https://jestjs.io) as testing framework
* runs tests as part of build


t8 (changes requested):
Thanks so much for the contribution, @arein! We use `yarn` instead of `npm`, so would you mind removing `package-lock.json` here?

martonlederer:
Nice work :). Any chance you could creare a GitHub workflow for this?

arein:
> Nice work :). Any chance you could creare a GitHub workflow for this?

It's already covered by the [.github/workflows/build.yml](build workflow) now that `yarn build` also runs the tests.

I think that's good for now. A dedicated workflow should also capture the test coverage but I think that warrants a dedicated Pull Request.

martonlederer:
Hmm I guess you're right. Merging this now :)

PR #28, opened 2021-07-09
Fix: Transfers over the limit cannot be made via `window.arweaveWallet.sign`.

Xaber20110202:
Fixes.
1. beyond 0.1 AR, cancel, cannot get error message; even if the password is entered correctly, the transfer cannot be made.
2. `window.addEventListener("beforeunload", cancel);` causes the cancel function to be called even if the password is entered correctly and the correct result is transmitted, resulting in the transfer never being completed


martonlederer (approved):
LGTM

PR #46, opened 2021-10-26
Make it possible for multiple wallet operations to process in parallel

DanMacDonald:
Currently in ArConnect for each requested operation the injected code posts a message to the plugin an adds an EventListener to await the result. This listener removes itself when it receives a `<operation>_result` message from the plugin and resolves the client call with the data payload.

Where this becomes a problem is when you have several long running operations at once.

If the client application makes three `decrypt` requests all at once, passing three different pieces of data to ArConnect to decrypt, ArConnect will post those message to the window and add a listener for each one.  So far so good. That is until the first `decrypt_result` message comes back from the plugin. At this point all three `decrypt` listeners receive the `decrypt_result` message and return the same data to the client. 

The expected behavior would be for each listener to only resolve when the `decrypt_result` matched with its `decrypt` request arrives. To achieve this behavior, this PR adds an `id` property to each message request, and listeners look for those `id`'s in the `<operation>_result` messages before returning the data to the client application.


src/scripts/background.ts
```diff
@@ -318,7 +335,8 @@ const handleApiCalls = async (
         type: "sign_transaction_result",
         ext: "arconnect",
         sender: "background",
-        res: true
+        res: true,
+        id: message.id
         //...(await signTransaction(message, tabURL))
```
DanMacDonald, line 340: This seems less than ideal? Hopefully just WIP?

t8:
Yes, @martonlederer is working on making some changes to fix a transaction signing bug inside of ArConnect on Chromium. Stay tuned!

martonlederer:
Amazing, will test this and merge later today

src/utils/injected.ts
```diff
@@ -59,12 +59,19 @@ export function createCoinWithAnimation() {
   }, 100);
 }
 
+var messageId = 0;
```
martonlederer, line 62: Could you change this to `let`?

src/scripts/content.ts
```diff
@@ -33,7 +33,11 @@ const connection = browser.runtime.connect(browser.runtime.id, {
 // forward messages from the api to the background script
 window.addEventListener("message", async (e) => {
   if (!validateMessage(e.data, {}) || !e.data.type) return;
+  let id = e.data.id;
```
martonlederer, line 36: We don't need a variable for this, let's remove it

src/utils/injected.ts
```diff
@@ -59,12 +59,19 @@ export function createCoinWithAnimation() {
...
+
 export const callAPI = (message: Omit<MessageFormat, "ext" | "sender">) =>
   new Promise<void | any>((resolve, reject) => {
+    // give every message a unique autoincrementing id
+    let id = messageId;
```
martonlederer, line 67: We won't change this, let's use a `const` here

martonlederer (changes requested):

src/utils/injected.ts
DanMacDonald: yep 👍

src/scripts/content.ts
DanMacDonald: done!

src/utils/injected.ts
DanMacDonald: good idea

martonlederer (approved):
LGTM

PR #53, opened 2021-11-18
feat: add Ledger wallet integration

CDDelta:
This PR adds support for connecting ArConnect to a Ledger with an Arweave wallet.

Note:
- Multiple Ledger wallets can be added but only one may be active (connected to the browser) at once.
- If the active wallet in ArConnect does not match the one on the connected Ledger an error is thrown.
- Config exports exclude Ledger wallet records.
- Due to limitations with the Ledger x Arweave API, encryption/decryption and custom signature requests resolve with an "Action not supported" error when called.
- Transaction signing requests to a Ledger will error if provided with custom signature options.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@primer/octicons-react": "^16.0.0",
     "@verto/js": "^0.0.0-alpha.19",
     "@verto/lib": "^0.10.4",
-    "@zondax/ledger-arweave": "^0.2.0",
+    "@zondax/ledger-arweave": "^1.0.0",
     "ar-gql": "^0.0.6",
     "arverify": "^0.0.11",
     "arweave": "^1.10.18",
--- a/package.json
+++ b/package.json
@@ -14,10 +14,12 @@
   },
   "dependencies": {
     "@geist-ui/react": "^2.2.0",
+    "@ledgerhq/hw-transport-webusb": "^6.11.2",
     "@limestonefi/api": "^0.1.3",
     "@primer/octicons-react": "^16.0.0",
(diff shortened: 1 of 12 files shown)


CDDelta:
Blocked on some typings issue with the [Ledger SDK](https://github.com/Zondax/ledger-arweave/pull/43). We could potentially release with our own fork release of the SDK if needed.

martonlederer:
Is this ready to merge or should I merge the other one? I believe that also has the ledger stuff

CDDelta:
Both are not ready yet, was pending a release of the Ledger SDK from Zondax's side which just got updated, will get on this tomorrow. When the time comes this will have to be merged first.

martonlederer:
Alright, thank you! Make sure to tag me when it's time

CDDelta:
@martonlederer this PR is ready for your review. It conflicts with a change on `development` but I'm not sure which way to resolve it. What was the reasoning behind using `getUploader()` as opposed to `post()`?

martonlederer:
@CDDelta I noticed that the network is more likely to drop txs when it's using "post". It even worked better for me, than fee multiplication

CDDelta:
> @CDDelta I noticed that the network is more likely to drop txs when it's using "post". It even worked better for me, than fee multiplication

Interesting, resolved the merge conflicts with that and added a comment for it.

martonlederer:
Thanks! No idea why that happens...

heiningair:
Is this feature request still in the making? If not is there an estimate when this could work, since ledger nano s already integrates arweave?

PR #65, opened 2022-03-07
feat: Link to viewblock for wallet address

hatskier:
This PR converts a simple plain text with user address to a link that opens the address details on the viewblock webpage.

### DIFF
--- a/src/components/WalletManager.tsx
+++ b/src/components/WalletManager.tsx
@@ -78,6 +78,10 @@ export default function WalletManager() {
     setVerifiedAddresses(loaded);
   }
 
+  function getViewblockLinkForAddress(address: string) {
+    return `https://viewblock.io/arweave/address/${address}`;
+  }
+
   function deleteWallet(addr: string) {
     if (wallets.length - 1 > 0) {
       if (profile === addr)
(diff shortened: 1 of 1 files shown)


src/components/WalletManager.tsx
```diff
@@ -121,7 +125,12 @@ export default function WalletManager() {
           {verifiedAddresses.includes(profile) && <VerifiedIcon />}
         </h1>
         <p className={styles.Address}>
-          {formatAddress(profile)}
+          <Tooltip text="Check address on viewblock">
+            <a href={getViewblockLinkForAddress(profile)} target="_blank">
```
martonlederer, line 129: Are you sure this is going to work? Hrefs don't work in popups, so I usually do an onclick event and open a new tab using the chrome API.

martonlederer (changes requested):

src/components/WalletManager.tsx
martonlederer: Nevermind this is fine

martonlederer, line 129: This changes the color of the address. The color should be inheriting the color of the parent `p` element.
Screenshot 2022-08-14 at PM 2 42 11


martonlederer (changes requested):

PR #72, opened 2022-05-04
Gateway update

martonlederer:
### The Gateway Update
This update includes additional features and changes to the existing gateway config system

Changelog:
- Renamed "Arweave Config" to "Gateway"
- Added Gateway Explorer
- Added suggested gateways
- Added gateway status to home screen
- Added gateway config API for developers: now developers can request users to use a custom gateway on connection
- Added gateway storage: each app has it's own gateway configuration for signing, etc.
- Moved all resources to be fetched from the configured gateway


martonlederer:
@t8 Please try this out, see how it works :)

PR #75, opened 2022-06-25
replace docs link with correct url

PBillingsby:
The current docs link is pointed at `https://www.arconnect.io/docs` which is a 404. This PR replaces it with the most recent docs at `https://docs.th8ta.org/arconnect`


martonlederer:
Thanks!

PR #76, opened 2022-07-06
fix(archive integration): Fetch drives from all ArFS compatible transactions

fedellen:
This PR removes the `"App-Name"` GQL tag filter which will enable the ArConnect extension to use the archive feature to any ArFS compatible drive, including the ArDrive-CLI and apps using ArDrive-Core.

The ArDrive apps use more `App-Name`s than just `"ArDrive-Web"` and `"ArDrive-Desktop"`. We currently also use `"ArDrive-CLI"`, `"ArDrive-Core"`, and `"ArDrive-Sync"`. But adding a query with this many tag values has proven to be intensive on the gateway. 

The ArFS spec is not really limited by App-Name. All MetaData Tx can be queried by `ArFS` with value `0.10` or `0.11`. But removing the App-Name tag filter altogether works just as well here and is more future proof

### DIFF
--- a/src/views/Archive/App.tsx
+++ b/src/views/Archive/App.tsx
@@ -301,7 +301,6 @@ export default function App() {
             first: 100
             owners: [$addr]
             tags: [
-              { name: "App-Name", values: ["ArDrive-Desktop", "ArDrive-Web"] }
               { name: "Entity-Type", values: "drive" }
               { name: "Drive-Privacy", values: "public" } 
             ]


martonlederer:
Thanks for the pr! Wouldn't it make sense to use ArConnect, as the value of the "App-Name" tag?

fedellen:
This query is being used to fetch the public ArFS drives of a given wallet address. Using Arconnect tag filter with drive and public would probably net 0 results. There are no tags being added here

If that question was meant to be directed towards when we write new ArFS transactions, e.g archive content to ArDrive: I would say it's fine to use Arconnect as the App-Name there. But you're already using ArDrive-Client: Arconnect which is a much more performance query. App-Name is a particularly congested name at the gateway 🤓

fedellen:
Any update here? Do you need anything else from me?

martonlederer:
Sorry it should be perfect

PR #78, opened 2022-08-14
Speed up transaction signing

martonlederer:
- Removes fee tx creation from the tx signing API
- Creates fees using the [Alarms API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms) (added an another permission to the manifest)
- Uses redstone api for fetching AR price

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
   },
   "dependencies": {
     "@geist-ui/react": "^2.2.0",
-    "@limestonefi/api": "^0.1.3",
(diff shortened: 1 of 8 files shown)


t8 (approved):
Amazing work @martonlederer. This is awesome.

In testing with Permanotes, the time to post a tx in ArConnect goes from ~34.49s --> ~3.45s

Averaging over a 10x speed improvement here 🎉

PR #85, opened 2022-09-27
refactor: add arlocal SuggestedGateway

mogulx-operates:
Hello! Just adding a way to select arlocal from the "Suggested Gateways".  Not sure if that's something you wanted in there but thought I'd put in a PR for it and see what happens.


martonlederer (approved):
Looking good, thank you :)

PR #91, opened 2022-10-23
Add non-breaking standard RSA decryption

### LINKED ISSUE #90: encrypt() and decrypt() APIs use own ArConnect-specific encryption methods
Is there any good reasons why ArConnect doesn't allow direct RSA encryption/decryption, without ArConnect's algorithm with adding symmetric keys there?

angrymouse:
Closes #90, detects what decryption way app wants to use (standard RSA one, like arweave.app uses, or old ArConnect one with AES in the middle), and uses it.


angrymouse:
@martonlederer Can you review?

martonlederer (approved):
LGTM

PR #93, opened 2022-12-10
fix: client can not to provide authentication credentials

dirname:
- Issue: 

https://user-images.githubusercontent.com/18101258/206866050-52cf68fa-20fe-4b49-9ee4-d8f4d8ce0b19.mov


martonlederer:
Hey there, sorry for the delay here. The next version of ArConnect will fix this issue, because the feature that causes this error has been removed.

Thank you for the PR :)

PR #96, opened 2023-03-26
fix: default arConfetti config to name instead of full url

dogukanakkaya:
In `src/api/modules/sign/utils.ts:29` icon already returned as joined with url, doing it again in the config was causing a duplication.

```js
return browser.runtime.getURL(`assets/${iconName}.png`);
```

Also fixes: https://github.com/arconnectio/ArConnect/issues/89

### DIFF
--- a/src/stores/reducers/settings.ts
+++ b/src/stores/reducers/settings.ts
@@ -17,7 +17,7 @@ export interface ISettingsAction {
 
 const defaultConfig: ISettings = {
   currency: "USD",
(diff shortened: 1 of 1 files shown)


dogukanakkaya:
Why is this closed without any mentions? I assume this is still an issue when I look at the codebase, no?

martonlederer:
We've released our 1.0 version, so this pull request is no longer relevant. If you still find this issue with the new version or in the codebase, please open a new issue/pr. Thanks!

PR #104, opened 2023-07-29
feat: adds update password upon expiration

nicholaswma:
Feature adds updatePassword function that updates with a new password, and UI changes for when a password is expired
image

### DIFF
--- a/src/wallets/index.ts
+++ b/src/wallets/index.ts
@@ -1,4 +1,9 @@
-import { addExpiration, checkPassword, getDecryptionKey } from "./auth";
+import {
+  addExpiration,
+  checkPassword,
+  getDecryptionKey,
+  setDecryptionKey
+} from "./auth";
 import { decryptWallet, encryptWallet } from "./encryption";
 import type { JWKInterface } from "arweave/node/lib/wallet";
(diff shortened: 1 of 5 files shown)


martonlederer (changes requested):
Looks great! Just a few changed and it should be good to go!

src/wallets/index.ts
```diff
@@ -260,6 +265,34 @@ export async function addWallet(
...
+    throw new Error("Invalid password");
+  }
+  const wallets = await getWallets();
+  for (const item of wallets) {
+    const decrypedKeyfile = await decryptWallet(
+      (item as LocalWallet).keyfile,
+      prevPassword
+    );
+    const encrypted = await encryptWallet(decrypedKeyfile, newPassword);
+
```
martonlederer, line 288: After you've encrypted the wallet, the decrypted instance is no longer needed, so you should immediately clear it from the runtime memory, just as a security measure. We have the `freeDecryptedWallet()` utility function for this (from: `src/wallets/encryption.ts`).

Int this case, it would look something like this:
```ts
// remove wallet from runtime memory
freeDecryptedWallet(decrypedKeyfile);
```

martonlederer (approved):
All good now!

PR #106, opened 2023-08-02
feat: adds Segment (analytics-next) 

nicholaswma:
Adds Segment Analytics to track 
- If a wallet is funded
- Last login time
- Connected to external application
- Onboarding completed

- Adds setting tab to opt-out of analytics

Edit 8/21:
- also includes opt-in for existing users without 'setting_analytic' in their storage:
image
- and additional step for onboarding and existing users
image

All analytics recorded are anonymous

TODO: Add PLASMO_PUBLIC_SEGMENT_WRITEKEY to .env

https://www.npmjs.com/package/@segment/analytics-next


martonlederer (changes requested):
I need to update a few things, but once we have that this LGTM.

src/utils/analytics.ts
```diff
@@ -0,0 +1,61 @@
...
+  // first we check if we are allowed to collect data
+  const enabled = await ExtensionStorage.get<boolean>("setting_analytics");
+
+  if (!enabled) return;
+
+  const ONE_HOUR_IN_MS = 3600000;
+
+  // TODO:login is tracked only once and compared to an hour period before logged as another Login event
+  if (eventName === EventType.LOGIN) {
+    const storageTime = await TempTransactionStorage.get(EventType.LOGIN);
```
martonlederer, line 26: Are you sure you wanted to use `TempTransactionStorage` here @nicholaswma? This resets frequently, sometimes after minutes. As far as I can see, you want it persisted for at least an hour? If so, than I'd recommend using either the extension local storage or `window.localStorage`.

src/utils/analytics.ts
```diff
@@ -0,0 +1,61 @@
...
+export enum EventType {
+  FUNDED = "FUNDED",
+  CONNECTED_APP = "CONNECTED_APP",
+  LOGIN = "LOGIN",
+  ONBOARDED = "ONBOARDED"
+}
+
+export const trackEvent = async (eventName: EventType, properties: any) => {
+  // first we check if we are allowed to collect data
+  const enabled = await ExtensionStorage.get<boolean>("setting_analytics");
```
martonlederer, line 18: I moved this here, so we don't accidentally track anything if it is not allowed.

src/utils/analytics.ts
```diff
@@ -0,0 +1,62 @@
...
+export enum EventType {
+  FUNDED = "FUNDED",
+  CONNECTED_APP = "CONNECTED_APP",
+  LOGIN = "LOGIN",
+  ONBOARDED = "ONBOARDED"
+}
+
+export const trackEvent = async (eventName: EventType, properties: any) => {
+  // first we check if we are allowed to collect data
+  const enabled = await getSetting("analytics").getValue();
```
martonlederer, line 18: You should use the `getSetting().getValue()` function for getting setting values.

src/components/popup/home/AnalyticsConsent.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { useStorage } from "@plasmohq/storage/hook";
+import { CloseIcon } from "@iconicicons/react";
+import type { Variants } from "framer-motion";
+import browser from "webextension-polyfill";
+import useSetting from "~settings/hook";
+import styled from "styled-components";
+
+export default function AnalyticsConsent() {
+  // store user answer
+  const [_, setAnalytics] = useSetting<boolean | undefined>("analytics");
```
martonlederer, line 12: We have the `useSetting` hook to get setting values in components.

src/utils/analytics.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!enabled) return;
+
+  // only track in prod
+  if (process.env.NODE_ENV === "development") return;
```
martonlederer, line 23: We only want to track in production.

martonlederer (approved):
Added everything I wanted, looks good now @nicholaswma. Feel free to merge :)

src/utils/analytics.ts
nicholaswma: Hmm, good point. I had the hour as a placeholder to not track every single time the user hits the main page but I think we can probably expand further on the specs; something like if a user opens the extension, closes it, and opens it again in the span of 10 seconds or so, do we consider that 2 logins?

nicholaswma:
@martonlederer One last thing, if a user doesn't interact with the opt-in analytics checkbox on the last step of the setup, it'll prompt the user with AnalyticsConsent when they first open the extension, is that by design?

martonlederer:
Not by design. Let me fix that quickly

martonlederer:
@nicholaswma this should fix it

PR #113, opened 2023-08-18
Allowance fixes

martonlederer:
- Fixes allowance limit check & reset
- Fixes allowance limit display (display in AR instead of winstons and make it more understandable)
- Fixes allowance for `signDataItem()` and `dispatch()`


nicholaswma (approved):
LGTM - logic looks great to me, only things I'd suggest are more UI/QOL and could be suggested later 

- Negative values shouldn't be allowed in limit input
- Password request on pop-up for allowance increase should either be removed or implemented into the settings page when changing the limit as well

PR #114, opened 2023-08-18
Token fixes

martonlederer:
- Add $U DRE node
- Update default DRE node
- Fix `addToken()` api errors
- Fix broken token balances on homepage
- Add support for `decimals` specs (link wip)
  - This replaces `divisibility`
  - Legacy tokens with the `divisibility` option will still be supported
  - `decimals` works similarly to how it is done on Ethereum: state stores integer balance in the smallest unit possible for the token, this balance is divided by `10^decimals` to get the actual, fractioned balance)
- Fix oversized collectible images not centered
- Added Bazar link for collectibles


t8:
@martonlederer here are the logs I get when attempting to add PIXL tokens to my ArConnect wallet:

image

Feel free to reproduce on this link by clicking the "ArConnect" button in the top right: https://viewblock.io/arweave/contract/tfalT8Z-88riNtoXdF5ldaBtmsfcSmbMqWLh2DHJIbg

martonlederer:
Really weird, I was able to reproduce it, but now the error is gone again

martonlederer:
Ready for review.

martonlederer:
Some of the functionalities I am implementing here could be added to ArweaveKit.

martonlederer:
Okay now this is ready

t8 (changes requested):
Adding the UCM contract now works, but PIXL tokens were considered collectibles rather than fungible tokens, and it displayed in my wallet after adding even though I didn't hold any.

I was also unable to see the specific balance of the collectible in my wallet. After manually switching to a token, it worked, but didn't render the icon.
image

**Some requested changes:**
- Can we switch the token balance text with the estimated value text (primary should be token balance secondary should be estimated dollar value)?
image

- Can we render up to 8 assets at a time in the main wallet view page?
image

- Could we detect the image icon for atomic fungible tokens automatically rather than defaulting to an empty icon?
image

- Can we have adding tokens from ViewBlock default to fungible tokens rather than collectibles, or alternatively have a standard we could use to detect this automatically?

The current fixes you've implemented seem to work; just thinking of ways to improve UX further.

t8 (approved):
Awesome stuff. LGTM

PR #115, opened 2023-08-20
fix(events): fixed app config emitter

atticusofsparta:


martonlederer (changes requested):

martonlederer (approved):
LGTM! Much needed fix, thank you for your help.

PR #116, opened 2023-08-22
`ar://` protocol fixes

martonlederer:
- Fix interception not working (using ar.io's updated method)
- Fix parser
- Fix path parsing (e.g.: `ar://bazar/favicon.svg` and other full urls work now)


nicholaswma (approved):
LGTM

PR #117, opened 2023-08-25
fix: Prevent useEffect multiple executions

pawanpaudel93:
# Fix

- Prevent multiple retriggers of the `useEffect` hook due to the use of `setWallets` inside useEffect and `wallets` being the dependency.


martonlederer (approved):
LGTM, merge when you want.

PR #121, opened 2023-09-05
Permission fixes

martonlederer:
- Fixes additional permissions not being added for the application
- Removes the error thrown when requesting for permissions that the app already has


pawanpaudel93:
- The first scenario works as expected.
- The second scenario - `TODO`
- The third scenario works and does not throw an error when all or some granted permissions are again requested.
- The fourth scenario works and does not throw an error on requesting some already granted  + some new permissions.
If firstly connected with permissions `['SIGN_TRANSACTION', 'ACCESS_PUBLIC_KEY', 'SIGNATURE']`, then secondly connected with permissions `['SIGNATURE', 'DISPATCH']`. The final permissions will be `['SIGNATURE', 'DISPATCH']`.

pawanpaudel93:
@martonlederer Yes, the second and the fourth don't work as expected as the current permissions are not shown during the new connection popup but only the requested ones (mix of already granted + ungranted) are shown.

pawanpaudel93:
@martonlederer Now all the above test scenarios are working as expected after the recent commit.

pawanpaudel93 (approved):

PR #123, opened 2023-09-18
Encryption and decryption fix

martonlederer:
- Fixes new encryption and decryption implementation
  - Fixes type assertions
  - Fixes private and public key import
  - Fixes missing finalizer to transform back to buffer
- Fixes old encryption implementation
  - Support encrypting string inputs
  - Fix buffer inputs

### DIFF
--- a/src/api/foreground.ts
+++ b/src/api/foreground.ts
@@ -36,7 +36,9 @@ import encrypt, {
   finalizer as encryptFinalizer
 } from "./modules/encrypt/encrypt.foreground";
 import decryptModule from "./modules/decrypt";
-import decrypt from "./modules/decrypt/decrypt.foreground";
+import decrypt, {
+  finalizer as decryptFinalizer
+} from "./modules/decrypt/decrypt.foreground";
 import signatureModule from "./modules/signature";
 import signature, {
   finalizer as signatureFinalizer
@@ -69,7 +71,7 @@ const modules: ForegroundModule[] = [
   { ...signModule, function: sign, finalizer: signFinalizer },
   { ...dispatchModule, function: dispatch, finalizer: dispatchFinalizer },
   { ...encryptModule, function: encrypt, finalizer: encryptFinalizer },
-  { ...decryptModule, function: decrypt },
+  { ...decryptModule, function: decrypt, finalizer: decryptFinalizer },
   { ...signatureModule, function: signature, finalizer: signatureFinalizer },
   { ...addTokenModule, function: addToken },
   { ...isTokenAddedModule, function: isTokenAdded },
--- a/src/api/modules/decrypt/decrypt.background.ts
+++ b/src/api/modules/decrypt/decrypt.background.ts
@@ -90,13 +90,16 @@ const background: ModuleFunction<string | Uint8Array> = async (
       options.salt
     );
 
-    // if a salt is present, split it from the decrypted string
-    if (options.salt) {
-      return arweave.utils.bufferToString(res).split(options.salt)[0];
-    }
-
     // remove wallet from memory
     freeDecryptedWallet(privateKey);
+
+    // if a salt is present, split it from the decrypted string
+    if (options.salt) {
+      const rawSalt = new TextEncoder().encode(options.salt);
+
+      // TODO: see why this doesn't work
+      return res.slice(0, res.length - rawSalt.length);
+    }
 
     return res;
   } else if (options.name) {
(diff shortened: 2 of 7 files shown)


pawanpaudel93:
Hey @martonlederer, on using the old deprecated way with salt, the decryption is having issues here as it is not returning the correct output. For example:

```ts
await window.arweaveWallet.connect(["ENCRYPT", "DECRYPT"]);

// encrypt data
const encrypted = await window.arweaveWallet.encrypt(
  "This message will be encrypted",
  {
    algorithm: "RSA-OAEP",
    hash: "SHA-256",
    salt: "hero"
  }
);

console.log("Encrypted bytes:", encrypted);

// decrypt data
const decrypted = await window.arweaveWallet.decrypt(
  encrypted,
  {
    algorithm: "RSA-OAEP",
    hash: "SHA-256",
    salt: "hero"
  }
);

console.log("Decrypted data:", new TextDecoder().decode(decrypted));
```

I am getting `decrypted` as a uint8Array having length 30 with all its elements as 0.
 [https://github.com/arconnectio/ArConnect/blob/e8314be1f4a8a9a39d05791fde0c5698b5eab467/src/api/modules/decrypt/decrypt.background.ts#L94](https://github.com/arconnectio/ArConnect/blob/e8314be1f4a8a9a39d05791fde0c5698b5eab467/src/api/modules/decrypt/decrypt.background.ts#L94) 

Maybe using something like this or similar code with help with the issue.
```ts
function removeSaltFromUint8Array(originalArray: Uint8Array, salt?: string) {
  if (!salt) return originalArray;

  // Convert the string to a Uint8Array
  const stringArray = new TextEncoder().encode(salt);

  // Calculate the start index to remove the string from the end
  const startIndex = originalArray.length - stringArray.length;

  // Check if the start index is valid and the substring at that position matches the salt
  if (
    startIndex >= 0 &&
    originalArray
      .subarray(startIndex)
      .every((byte, index) => byte === stringArray[index])
  ) {
    // Create a new Uint8Array that excludes the bytes corresponding to the string
    const newArray = new Uint8Array(startIndex);

    // Copy bytes before the string
    newArray.set(originalArray.subarray(0, startIndex));

    return newArray;
  } else {
    // Return the original array if the string is not found at the end
    return originalArray;
  }
}
```

pawanpaudel93:
Passing data as Uint8Array works but when ArrayBuffer is passed it does not work.

```ts
// connect to the extension
await window.arweaveWallet.connect(["ENCRYPT", "DECRYPT"]);

// encrypt data using RSA-OAEP
const encrypted = await arweaveWallet.encrypt(
    new TextEncoder().encode("This message will be encrypted").buffer,
    { name: "RSA-OAEP" }
);

console.log("Encrypted bytes:", encrypted);

// now decrypt the same data using
// the same algorithm
const decrypted = await arweaveWallet.decrypt(
    encrypted,
    { name: "RSA-OAEP" }
);

console.log(
    "Decrypted data:",
    new TextDecoder().decode(decrypted)
);
```

pawanpaudel93:
As the new approach doesn't use the hybrid approach (RSA + AES), it can only encrypt/decrypt smaller data. The following code will not work and throw an error. Maybe this should be communicated in the docs if this is the expected behavior that this is useful to encrypt/decrypt smaller data only. On the other hand, the deprecated old method works as it uses the hybrid approach (RSA + AES).

```ts
await window.arweaveWallet.connect(["ENCRYPT", "DECRYPT"]);

const dataToEncrypt = new TextEncoder().encode(
  "ArConnect is an Arweave-native wallet extension that provides a non-custodial wallet and asset management in your favorite browser. The extension allows Arweave wallet holders to interact with any dApps without sharing their private keys with them." +
    "The isolated environment that ArConnect creates is not only a security improvement for users, but it also provides a more seamless login flow for applications. Developers no longer have to build sign-in functionality, they can let ArConnect do the hard work for them."
);

console.log("dataToEncrypt: ", dataToEncrypt);

const encrypted = await window.arweaveWallet.encrypt(
  new Uint8Array(dataToEncrypt),
  {
    name: "RSA-OAEP"
  }
);

console.log("Encrypted bytes: ", encrypted);

// now decrypt the same data using
// the same algorithm
const decrypted = await window.arweaveWallet.decrypt(encrypted, {
  name: "RSA-OAEP"
});

console.log("Decrypted data:", decrypted);
```

martonlederer:
@pawanpaudel93 these should be fixed now. Please check again.

pawanpaudel93:
@martonlederer LGTM 🚀

pawanpaudel93 (approved):

PR #125, opened 2023-09-20
fix: adds opt in to generate page, remove logs

nicholaswma:
- Adds opt-in analytics option on wallet generation


martonlederer (approved):
LGTM

PR #138, opened 2023-10-19
Seedphrase improvements

martonlederer:
- Don't show head for seedphrase confirmation
- Pre fill random words for easier verification of saved seedphrase


nicholaswma (approved):
LGTM

PR #147, opened 2023-11-08
feat: track total AR - alarm

nicholaswma:
Revised method to track total AR on ArConnect.

On install will initialize
- an alarm to run every 30 days
- uuid for a user (we can't rely on anonymous id here due to service worker limitations)


t8 (approved):
LGTM! Nice work @nicholaswma.

PR #149, opened 2023-11-10
feat: adds anonymous tracking on sign

nicholaswma:
- Adds anonymous tracking when sign is called
- Records appUrl and transaction total


t8 (approved):
LGTM. Nice work here @nicholaswma

PR #152, opened 2023-11-16
Buy Buttons - Dashboard (w/ balance & w/ no balance)

elehma4:
This PR encompasses both ARC-208 & ARC-209 including the new Buy AR buttons. On the no balance dashboard it also includes the newly updated design with the new Transfer AR button beneath the Buy AR button. The Buy AR button is a new, reusable component.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,5 +40,3 @@ keys.json
 
 # temporary remove development icon
 assets/icon.development.png
-
-notes.txt
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,5 @@ keys.json
 
 # temporary remove development icon
 assets/icon.development.png
+
+notes.txt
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -39,4 +39,6 @@ keys.json
 .tsbuildinfo
 
 # temporary remove development icon
-assets/icon.development.png
\ No newline at end of file
+assets/icon.development.png
+
+notes.txt
```
nicholaswma: Don't think this is necessary, unless there's reason

nicholaswma (changes requested):
Nice work! I'd just remove the addition to gitignore since it's not relevant to the other developers. I want to move BuyButton to @arconnect/components - but that might be dependent on the agreed design.

.gitignore
elehma4: Updated .gitignore to remove notes.txt

PR #155, opened 2023-11-20
feat: triggers ar track on install, removed logic for uuid creation

nicholaswma:
triggers AR track on install, removed logic for uuid creation


7i7o (approved):
Everything looks good! 🚀

PR #156, opened 2023-11-21
fix: temp disables u-token on send page

nicholaswma:
Temporarily disables U-token on send page due to transfer bug

image


7i7o (approved):
🚀

PR #157, opened 2023-11-28
ARC-210 Buy AR Screen (Live with API)

elehma4:
This branch includes solutions for live fiat to AR conversions, conversion edge cases with error handling, and saving quote data to local storage.


7i7o (changes requested):
It's getting quotes from the API and that is great! 💪 

A few things to consider:
- Amounts with a final dot show "invalid conversion" (could be solved by sanitizing the input). This warning is also shown when you need to put decimals and you didn't finish typing the amount.
- There's no feedback to show it's getting a quote, showing the outdated quote while it's fetching, or even when the amount is invalid.
- Sometimes API calls return in an unordered fashion and you get the wrong quote (see video)

https://github.com/arconnectio/ArConnect/assets/84824996/31293090-b3b6-44c7-9487-94f675477ba5

- There's a maximum amount, and that case is missinformed as "increase fiat amount"

Changes still pending:
- ~~Light Mode~~ -> Done

7i7o (approved):
Working right now. Bugs can be addressed in later tickets 👍

PR #161, opened 2023-11-30
ARC-211 Confirm Purchase

elehma4:
This PR includes the new confirm purchase page, after the user requests a converion of fiat into AR. It shows the user all of the order details and when the user decides to go through with the purchase it then sends their order data to the onramp and re-routes the user to begin KYC.


7i7o (approved):
LGTM! 🚀

PR #163, opened 2023-12-01
fixup: no balance dashboard

elehma4:
This PR cleans up the no balance dashboard. It fixes the size of the art component so that everything fits in the screen. It also fixes the mismatched sizes of the Buy AR button and the Transfer button. As well as fixes the hover effect on the transfer button.


7i7o (approved):
It shows great in the available space now! 🙌

PR #164, opened 2023-12-04
ARC-217 payment method dropdown menu

elehma4:
This PR includes the new select input/modal pop-up for payment methods. This component will be used for fiat selections as well in the next PR (ARC-218).


src/components/InputMenu.tsx
7i7o, line 1: It works 👍 
Since this is a new component, it would be nice to be able to have a generic InputMenu so we don't have to copy the component and modify it with the options when we need to re-use this one for currencies or any other case.

7i7o (approved):
Approved 🚀 

We need to refactor InputMenu for currency selection and any new use cases

PR #165, opened 2023-12-05
ARC-218 fiat currency dropdown menu

elehma4:
This PR includes the updated fiat select & search menu. It also modularizes the InputMenu component to be used for both the payment methods and fiat currencies.


7i7o (approved):
tACK! LGTM! 🚀

PR #166, opened 2023-12-06
ARC-220 feedback fetching quotes

elehma4:
This PR adds a loading indicator to give the user feedback when a new quote is being fetched.

### DIFF
--- a/src/routes/popup/purchase.tsx
+++ b/src/routes/popup/purchase.tsx
@@ -6,7 +6,14 @@ import { useStorage } from "@plasmohq/storage/hook";
 import { ExtensionStorage } from "~utils/storage";
 import { useTheme, hoverEffect } from "~utils/theme";
 import { CloseIcon, ChevronDownIcon } from "@iconicicons/react";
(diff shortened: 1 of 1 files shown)


src/routes/popup/purchase.tsx
```diff
@@ -6,7 +6,14 @@ import { useStorage } from "@plasmohq/storage/hook";
 import { ExtensionStorage } from "~utils/storage";
 import { useTheme, hoverEffect } from "~utils/theme";
 import { CloseIcon, ChevronDownIcon } from "@iconicicons/react";
-import { Section, Card, Spacer, Button, Select } from "@arconnect/components";
+import {
+  Section,
+  Card,
+  Spacer,
+  Button,
+  Select,
```
7i7o: declared but never read

7i7o (approved):
LGTM! 🚀 

We still need to create a small delay from when the user stops typing until we fetchthe quote, so as not to spam with quotes for every keystroke of the user

PR #167, opened 2023-12-07
ARC-222 improve error handling

elehma4:
This PR includes the error message that details how much fiat a user needs for a successful quote. Example: 'Amount should be in between EUR 28 and EUR 5575.' It also fixes a small styling bug that was happening when the loading icon renders when a new quote is fetched. It also includes a 200ms delay for each fetch.


7i7o (approved):
LGTM! 🚀

PR #169, opened 2023-12-07
ARC-224 add payment method to confirmation screen

elehma4:
This PR adds Payment Method to the user's order details page.


src/routes/popup/confirm.tsx
```diff
@@ -50,6 +51,12 @@ export default function ConfirmPurchase() {
       setSelectedFiat(quote.selectedFiat.toUpperCase());
       setFiatAmount(quote.fiatAmount);
       setPaymentMethod(quote.selectedPaymentMethod);
+      const paymentChoice = quote.availablePaymentMethods.find(
+        (method) => method.paymentTypeId === quote.selectedPaymentMethod
+      );
+      if (paymentChoice) {
```
nicholaswma, line 57: is there an instance where paymentchoice would be undefined?

nicholaswma:
LGTM

PR #170, opened 2023-12-11
ARC-225 retain quote data on back from confirm

elehma4:
This PR adds a new variable to local storage to handle if the user has gone back on confirming their purchase. It retains the existing data so that the user does not have to input their previous data. The logic takes place when component mounts.


nicholaswma:
Lgtm ✅

PR #171, opened 2023-12-11
fix: re-enable u-token transfer

### LINKED ISSUE #133: U transfers don't work?
Originally mentioned here, but will start a new issue: https://github.com/arconnectio/ArConnect/issues/128#issuecomment-1746816556

Neither my first or 2nd transaction resulted in a successful U transfer, but the UI gives confirmation of success and a tx.  My balance does not decrease, and the recipient's does not increase. 

1st tx:  https://viewblock.io/arweave/tx/DZ4LH5Udq4pAaban8gXhJEDBlCuvXL-bTA_ggsHZ6XA 

2nd tx:  https://viewblock.io/arweave/tx/6L-AzvLM3FJdpn2bfKFkp2uKxLRG13hdhUoAFVbYWDI

nicholaswma:
This PR re-enables U-Token Transfer, using `sendRequest` to send an HTTP request directly to warp 

- Message field is currently hidden on u-token transfer 
- Post-transaction, users are redirected to the home page of the wallet rather than the post-transaction screen, this is due to having to use a different endpoint to fetch tx (WIP)

### DIFF
--- a/src/routes/popup/send/auth.tsx
+++ b/src/routes/popup/send/auth.tsx
@@ -33,9 +33,11 @@ import browser from "webextension-polyfill";
 import Head from "~components/popup/Head";
 import styled from "styled-components";
 import Arweave from "arweave";
-import { defaultGateway, Gateway } from "~gateways/gateway";
-
-export default function SendAuth() {
+import { defaultGateway, type Gateway } from "~gateways/gateway";
+interface Props {
+  tokenID?: string;
+}
+export default function SendAuth({ tokenID }: Props) {
   // loading
   const [loading, setLoading] = useState(false);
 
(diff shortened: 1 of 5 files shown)


src/routes/popup/send/auth.tsx
```diff
@@ -259,7 +285,11 @@ export default function SendAuth() {
           content: browser.i18n.getMessage("sent_tx"),
           duration: 2000
         });
-        push(`/transaction/${transaction.id}?back=${encodeURIComponent("/")}`);
+        uToken
+          ? push(
+              `/transaction/${transaction.id}?back=${encodeURIComponent("/")}`
+            )
+          : push("/");
```
7i7o: Why is this one the opposite of the one in lines 186-190? 🤔

7i7o (approved):
LGTM! 🚀 
(Worked, but only tested with Test-U-Token)

7i7o (approved):
LGTM! 🚀

PR #172, opened 2023-12-13
fix: fixes generic error messaging with invalid tags

nicholaswma:
fix: fixes generic error messaging with invalid tags


7i7o (approved):
utACK! LGTM! 🚀

PR #173, opened 2023-12-13
ARC-230 ensure all available payment methods

elehma4:
This PR provides an update to the InputMenu component when handling payment methods. It now requires the selectedFiatCurrency and fetches then renders the available payment methods from the API.


7i7o (approved):
Working 💯 %!
LGTM! 🚀

PR #176, opened 2023-12-14
ARC-219 handle decimals for non english regions

elehma4:


7i7o (approved):
Works great! tACK
LGTM! 🚀

PR #180, opened 2023-12-19
ARC-233 no pricing data on dashboard

elehma4:
- Added useStorage hook to persistently store the last known AR price.
- Modified useEffect in Balance.tsx to use the stored price as a fallback when the API fails to fetch the current price.
- Ensured the stored price is updated with each successful API response.
- This fix ensures continuous display of fiat balance, reducing dependency on API availability.


7i7o (approved):
utACK. LGTM! 🚀

PR #185, opened 2023-12-22
fix: hardcode fallback gateway

nicholaswma:
- if there's no response from posting to arweave.net w/in 10 seconds, will fallback to post to ar-io.net


7i7o (approved):
LGTM! 🚀

PR #187, opened 2024-01-03
ARC-226 make confirm password on send optional

elehma4:
Updates:
- Created a new setting called Sign settings
- Existing Notify when signing a transaction setting functions as before
- New sub-setting called Password Required Allowance has been created with an initial value of 10
- The user can update the value, and that value will persist

Current UX:
- The user can click the Send button without having to input their password if the transaction amount is below the signature allowance.
- The user is required to input their password if the transaction amount is at or above the signature allowance.

UX Development to be continued:
- Send Screen 2 - Confirm transaction screen by @nicholaswma 
- Incorporates new UI to include the password input or not depending on the transaction amount vs. signature allowance amount.

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -391,6 +391,10 @@
     "message": "Sign notification",
     "description": "Sign notification settings title"
   },
+  "setting_sign_settings": {
+    "message": "Sign settings",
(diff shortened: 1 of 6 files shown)


7i7o (approved):
LGTM! 🚀

PR #189, opened 2024-01-08
feat: send screen page 1

nicholaswma:
https://communitylabs.atlassian.net/browse/ARC-237

NOTES: 
- General styling of inputs/buttons are within the send/index page, but will be moving those @arconnect/components before release
- There's an input scrollbar (number) that pops up on the amount input and the Input field border-radius does not match with the design ATM, this will be fixed with the update to the components package.

image


7i7o (approved):
tACK
New styling works perfect in dark mode
Light mode still has some white text that is almost unreadable

PR #190, opened 2024-01-10
fix: resets updatedToken state on address change

nicholaswma:
- Adds a useEffect to update token balances when wallet addresses change


7i7o (approved):
tACK
LGTM! 🚀

t8:
Can confirm this fixes the bug on my end!

PR #191, opened 2024-01-11
fix: disables ar balance to record on update, manually sets timer

nicholaswma:
- Previously, we've triggered an analytics call upon installation/update
- Removed that `await trackBalance();` on installation
- Manually updating alarms in Chrome (updates don't persist alarm states) and using storage to keep track of the monthly calls


7i7o (approved):
utACK
LGTM! 🚀

PR #193, opened 2024-01-12
feat: add analytics to AR Onramp

elehma4:


src/components/popup/home/BuyButton.tsx
```diff
@@ -23,10 +29,24 @@ export default function BuyButton({
...
+  };
 
   const targetRoute = route === "/purchase" ? "/purchase" : "/confirm-purchase";
 
-  const handleClick = () => {
+  const handleClick = async () => {
+    const eventType = eventMap[location];
+
+    if (eventType) {
+      console.log(eventType);
```
nicholaswma: remove

nicholaswma (commented):
utACK, one thing though, if we're only monitoring dashboard - purchase- confirm, we should consider using trackPage instead of trackEvent, wdyt?

nicholaswma (approved):
utACK lgtm!

PR #194, opened 2024-01-16
fix: monthly ar track set to beginning of the month

nicholaswma:
https://communitylabs.atlassian.net/browse/ARC-250

- Removed balance_tracker in ExtensionStorage since it's set to the beginning of the month now.
- Updated util functions


7i7o (approved):
utACK
LGTM! 🚀

PR #214, opened 2024-02-05
ARC-253 analytics to add

elehma4:


src/components/dashboard/SignSettings.tsx
```diff
@@ -37,12 +43,22 @@ export default function SignSettings() {
...
   };
 
-  const handleAllowanceChange = async (e) => {
+  const handleBlur = async (e) => {
     const newAllowance = Number(e.target.value);
-    setSignatureAllowance(newAllowance);
+    if (newAllowance !== signatureAllowance) {
+      trackEvent(EventType.SEND_ALLOWANCE_CHANGE, {
+        Before: signatureAllowance,
+        After: newAllowance
```
nicholaswma: before and after should be lowercase to follow camelcasing

src/routes/popup/send/confirm.tsx
```diff
@@ -334,6 +335,11 @@ export default function Confirm({ tokenID, qty }: Props) {
             content: browser.i18n.getMessage("sent_tx"),
             duration: 2000
           });
+          trackEvent(EventType.TX_SENT, {
+            contact: contact ? true : false,
+            amount: transactionAmount,
+            fee: networkFee
+          });
```
nicholaswma, line 342: we'll need to trackEvent on fallback case as well down on line ~400.
transaction amount from what i remember could be AR or other token types, If we're only monitoring AR transfers, we could have a check for `token === "AR" ? transactionAmount : 0`

nicholaswma (commented):
Added two minor points, but should be good after those!

7i7o (approved):
utACK 🚀

nicholaswma (approved):

PR #217, opened 2024-02-07
add background to expanded view

7i7o:
This PR adds a colored background to the expanded view, to make the extension more prominent in that state.

Example images:

Screenshot 2024-02-07 at 18 26 49
Screenshot 2024-02-07 at 18 27 29

### DIFF
--- a/src/components/popup/Route.tsx
+++ b/src/components/popup/Route.tsx
@@ -24,17 +24,21 @@ const Route: typeof BaseRoute = ({ path, component, children }) => {
   );
 };
 
-export const Wrapper = styled(motion.div)<{ responsive?: boolean }>`
+export const Wrapper = styled(motion.div)<{
+  responsive?: boolean;
+  expanded?: boolean;
+}>`
   position: relative;
   width: ${(props) => (props.responsive ? "100%" : "377.5px")};
-  min-height: 600px;
+  min-height: ${(props) => (props.expanded ? "100vh" : "600px")};
+  max-height: max-content;
+  background-color: rgb(${(props) => props.theme.background});
 `;
 
 const PageWrapper = styled(Wrapper)`
   position: absolute;
   top: 0;
   width: 100%;
-  background-color: rgb(${(props) => props.theme.background});
   transition: background-color 0.23s ease-in-out;
 `;
 
--- a/src/components/popup/Route.tsx
+++ b/src/components/popup/Route.tsx
@@ -84,12 +84,22 @@ const Page = ({
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/components/popup/Route.tsx
```diff
@@ -24,17 +24,21 @@ const Route: typeof BaseRoute = ({ path, component, children }) => {
...
+  min-height: ${(props) => (props.expanded ? "100vh" : "600px")};
+  max-height: max-content;
+  background-color: rgb(${(props) => props.theme.background});
 `;
 
 const PageWrapper = styled(Wrapper)`
   position: absolute;
   top: 0;
   width: 100%;
-  background-color: rgb(${(props) => props.theme.background});
```
nicholaswma, line 37: curious why this was moved to Wrapper

src/popup.tsx
```diff
@@ -30,87 +30,105 @@ import { NavigationBar } from "~components/popup/Navigation";
...
-            </HistoryProvider>
-          </Router>
-        </Wrapper>
+        <HideScrollbar expanded={expanded} />
+        <ExpandedViewWrapper>
+          <Wrapper expanded={expanded}>
+            <Router hook={useHashLocation}>
+              <HistoryProvider>
+                <Route path="/" component={Home} />
+                <Route path="/purchase" component={Purchase} />
```
nicholaswma, line 58: could we pass `expanded` to the pages to disable the variants/animations in Route.tsx?

nicholaswma (commented):
left some comments!

src/components/popup/Route.tsx
7i7o: Because in `PageWrapper` when filling bg in Receive/Send pages, it only filled up to the last element in them, not filling the whole height up to the navbar.

Try it out moving it back to page wrapper and you'll see the body bg in fullscreen

src/popup.tsx
7i7o: Sure!
I was actually wondering if we could change from a motion effect to a blur effect or something similar. At least in expanded view. Since I didn't have that much time, I pushed it as it was.

elehma4 (approved):
tACK 👌

src/popup.tsx
nicholaswma: sounds good! but yea, we can keep it minimal for now and look to implement another effect later on

7i7o: I pushed a small change to fix this. lmk if that works

nicholaswma (approved):
tACK, this looks great thanks! @7i7o

PR #220, opened 2024-02-08
Fix expanded view url on first open

7i7o:
This PR fixes a small bug that only happens after first installing the extension and clicking expanded view first, where the url would contain a `#` that would not let the full page view render.


nicholaswma (approved):
Lgtm!

PR #221, opened 2024-02-09
fix: ui issues and arns in search field

nicholaswma:
the screenshots are slightly confusing but the changes are that
- the "add" button is no longer disabled until a valid address is entered
- onClick, add will check if it's an address else check if it's an ArNS -> resolve to that address

the design choice was based on building on top of useMemo implementation for searching and searching ArNS names being an async fn

image

image

edit: added toast
image


elehma4 (approved):
tACK, LGTM 🙌

7i7o (approved):
tACK 🚀

PR #230, opened 2024-02-17
Use ArSvgies package + cache images in localStorage for wallets and contacts

7i7o:
This PR replaces the dynamic generative script with the @7i7o/arsvgies package to generate fallback avatars for wallets and contacts
This package includes an update to generate colors from a hash to clearly distinguish wallets that create a similar shape

It also stores the generated images in localStorage for wallets and contacts to avoid processing the generative art everytime a contact or wallet component in mounted in the UI

The PR unifies all the images shapes for wallets into squircles, removing the rounded images that were used in some places for contacts


nicholaswma (changes requested):
image

I think previously added contacts/wallets aren't getting svgies passed to them but other than that everything looks good!

7i7o:
> I think previously added contacts/wallets aren't getting svgies passed to them

You are right! I'm only generating the cached image for the ones being added 🤦 
I'll get to this after I finish with my other tasks 😊

PR #240, opened 2024-02-29
feat: added ability to paste process ids

nicholaswma:
Integrates ability to add tokens via process ids on settings page


elehma4 (approved):
tACK

PR #252, opened 2024-03-11
fix: dashboard ux/ui (arc-292)

elehma4:
- show tokens without an ar balance
- add loading icon to wallet balance
- fix display for no balance screen
- add useBalance hook


nicholaswma (approved):
tACK

PR #260, opened 2024-03-13
fix: notification denomination hotfix

nicholaswma:


nicholaswma:
combined with: https://github.com/arconnectio/ArConnect/pull/262

PR #262, opened 2024-03-13
fix: removed ao query for ticker

nicholaswma:


src/routes/popup/notifications.tsx
```diff
@@ -83,24 +89,66 @@ export default function Notifications() {
...
+        let ticker;
+        let quantityTransfered;
+        if (notification.isAo) {
+          // handle ao messages/sents/receives
+          let token = await fetchTokenByProcessId(notification.tokenId);
+          if (!token) {
+            ticker = formatAddress(notification.tokenId, 4);
+            quantityTransfered = notification.quantity;
+          } else {
+            ticker = token.Ticker;
```
7i7o, line 78: This should will throw an error when **no** "ao_tokens" are added added in settings.
`fetchTokenByProcessId` will return the `processId` because of `!tokens` and we will enter this `else` bracket with `token` being a string and not an object

src/routes/popup/notifications.tsx
```diff
@@ -83,24 +89,66 @@ export default function Notifications() {
...
+            );
+          }
+        } else if (notification.transactionType !== "Transaction") {
+          let token = await fetchTokenById(notification.tokenId);
+          if (!token) {
+            ticker = formatAddress(notification.tokenId, 5);
+            quantityTransfered = extractQuantityTransferred(
+              notification.node.tags
+            );
+          } else if (token.ticker !== "AR") {
```
7i7o, line 95: Same problem here if the user decides to remove all the warp tokens from settings.

7i7o (changes requested):
Maybe returning a falsy value in both `fetchTokenByProcessId` and `fetchTokenById`?

7i7o (approved):
tACK 🚀

PR #283, opened 2024-04-02
Arc 317 component v2 polishing

elehma4:
- update theme colors
- update send input buttons
- update icon button
- update input with button (arlocal)
- clean up imports
- rem units for a11y


7i7o (approved):
tACK 🚀

PR #286, opened 2024-04-04
arc-320 add vault to wallet selector drop down

elehma4:
- add vaults to wallet switcher and show vault icon


nicholaswma (approved):
tACK

PR #287, opened 2024-04-05
feat: added send flow from vault setting

nicholaswma:


src/components/dashboard/subsettings/VaultSettings.tsx
```diff
@@ -28,26 +38,35 @@ export default function VaultSettings({ initial = false, vaultName }: Props) {
...
             key={wallet.address}
             title={`${wallet.nickname} (${formatAddress(wallet.address, 3)})`}
             description={`${formatTokenBalance(balance)} AR`}
             active={false}
+            onClick={() => {
+              // Switch active accounts
+              setActiveAddress(wallet.address);
+              browser.tabs.create({
+                url: browser.runtime.getURL(
+                  `popup.html?expanded=true#/send/transfer/AR/${vaultAddress}`
```
7i7o, line 66: FYI: this is getting blocked in Arc Browser, complaining that the "popup" can only be opened in the actual popup and we should use one of the "tabs" routes to open extension urls in windows

7i7o (approved):
Opens up fine in Chrome and Brave

src/components/dashboard/subsettings/VaultSettings.tsx
nicholaswma: @7i7o good to know - we'll have to revisit this with the expanded view as well

PR #291, opened 2024-04-08
arc-325 vault wallet send receive page warning

elehma4:
- add warnings when transferring into or form a vault address

### DIFF
--- a/src/components/popup/Navigation.tsx
+++ b/src/components/popup/Navigation.tsx
@@ -14,12 +14,13 @@ import {
 } from "@untitled-ui/icons-react";
 import { ExtensionStorage } from "~utils/storage";
 import { useHistory } from "~utils/hash_router";
-import { useState, useEffect } from "react";
-import { getActiveAddress } from "~wallets";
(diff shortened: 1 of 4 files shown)


src/components/popup/Navigation.tsx
```diff
@@ -77,10 +128,45 @@ export const NavigationBar = () => {
...
+              <ButtonV2 secondary onClick={() => warningModal.setOpen(false)}>
+                {browser.i18n.getMessage("no")}
+              </ButtonV2>
+              <ButtonV2 onClick={() => confirmRouteChange()}>
+                {browser.i18n.getMessage("yes")}
+              </ButtonV2>
+            </>
+          }
+        >
+          <CenterText heading>Warning</CenterText>
```
nicholaswma: needs warning icon here

nicholaswma (changes requested):
- warning message font color should be white
- warning message is supposed to be different copy with send/receive
- font size is different w/ the design

nicholaswma (approved):

PR #295, opened 2024-04-10
fix: abbreviate large numbers

elehma4:
- adds M for millions, B for billions, T for trillions to tokens on homepage. 
- cleans up placeholder for no fiat value

### DIFF
--- a/src/components/popup/Token.tsx
+++ b/src/components/popup/Token.tsx
@@ -92,11 +92,11 @@ export default function Token({ onClick, ...props }: Props) {
       </LogoAndDetails>
       <BalanceSection>
         {isMillion ? (
-          <TooltipV2 content={totalBalance} position="left">
-            <NativeBalance>
+          <BalanceTooltip content={totalBalance} position="topEnd">
+            <NativeBalance style={{}}>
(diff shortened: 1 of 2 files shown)


7i7o (changes requested):
We are missing the tooltip to see the full/exact amount

7i7o (changes requested):
### Requested changes:

- Full amount is being rounded/floored after the 17th digit. Screenshot 2024-04-15 at 9 55 48 AM `1,234,567,891,234,567,891,234,567.89` should be the full amount in the tooltip


### Nice-to-haves:

- Tooltip should only be shown when we are abbreviating the number. Screenshot 2024-04-15 at 10 08 20 AM
Imagine you don't have any abbreviated balances in your wallet, what's the point of the tooltip in the UI?


- Moving Tooltip in the bottom/top because numbers `>99,999,999T` get cropped when it's on the left: Screenshot 2024-04-15 at 10 06 24 AM

7i7o (approved):
Approving without requested changes because we are migrating to use aoTokens library for token quantities

7i7o (approved):
tACK 🚀

PR #302, opened 2024-04-16
fix: sign allowance if 0

elehma4:


src/routes/popup/send/confirm.tsx
```diff
@@ -138,7 +142,7 @@ export default function Confirm({ tokenID, qty }: Props) {
 
     fetchData();
     trackPage(PageType.CONFIRM_SEND);
-  }, []);
+  }, [signAllowance, needsSign]);
```
7i7o:
```suggestion
  }, []);
```

You cannot add `signAllowance` and `needsSign` as dependencies of the useEffect because you are sanitizing (setting) them inside your useEffect code. Any update to them will trigger another call to useEffect creating an infinite loop.
That is why in the original code the dependencies don't include them so it is run only on component load

7i7o (dismissed):

nicholaswma (approved):
tACK
we can replace it with useStorage next  👍

PR #303, opened 2024-04-16
refactor: integrated ao-tokens + added ao logo to ao txns in notifications

nicholaswma:


elehma4 (approved):
tACK 🚀

src/tokens/aoTokens/ao.ts
```diff
@@ -157,10 +115,17 @@ export function useAoTokens(): [TokenInfoWithBalance[], boolean] {
...
+              const aoToken = await Token(id);
+              const balance = Number(
+                (await aoToken.getBalance(activeAddress)).toLocaleString()
+              );
+
+              return {
+                id,
+                balance
+              };
+            })
```
7i7o, line 128: We're inside a Promise.all, so if any one of the token Promises fails, everything fails.

Later on, we should look into a solution that resolves promises for tokens individually.

7i7o (approved):
tACK 🚀

PR #304, opened 2024-04-17
feat: support sending to ANS users

elehma4:


src/routes/popup/send/recipient.tsx
```diff
@@ -6,6 +6,7 @@ import {
 } from "~utils/storage";
 import {
   Input,
```
7i7o: Remove unused imports

7i7o (changes requested):
Code is 👍 

#### Changes needed:
- Add a Toast if you misspell or enter an ANS name that doesn't exist. When you enter a wrong arns, there's a Toast letting you know it didn't find that address.

- Add the correct placeholder in the Input "Address, **ANS** or ArNS name..."

7i7o (approved):
tACK 🚀

PR #305, opened 2024-04-18
feat: adds pop up for sign data item involving Transfer tag

nicholaswma:


src/routes/auth/signDataItem.tsx
```diff
@@ -0,0 +1,189 @@
...
+              description={formatAddress(recipient, 10)}
+              title={"Recipient"}
+              img={svgieAvatar}
+              onClick={() =>
+                browser.tabs.create({
+                  url: `https://viewblock.io/arweave/address/${recipient}`
+                })
+              }
+            />
+            {/* TODO: will we have access to process ID? if so would be great to get the token image and redirect to ao scanner */}
```
7i7o: We should have the `Target` and `Action` tag within that same message/DataItem to fetch the Token's Info

7i7o (approved):
tACK

PR #314, opened 2024-04-25
fix: send refactor

nicholaswma:
- Fixes an issue where PW was not being checking in the send flow
- Refactors sendAllowance to use `useStorage`


7i7o (approved):
tACK 🚀

PR #316, opened 2024-04-29
fix: update packages

nicholaswma:


package.json
```diff
@@ -121,6 +121,8 @@
   },
   "resolutions": {
     "arbundles/arweave": "^1.13.0",
+    "**/axios": ">=0.28.0",
+    "**/msgpackr": ">=1.10.1",
     "human-crypto-keys/node-forge": "^1.3.1",
     "human-crypto-keys/crypto-key-composer/node-forge": "^1.3.1",
     "@parcel/runtime-js": "2.8.3"
```
7i7o, line 128: The update now shows this warning for this version of `@parcel/runtime-js`:

```echo
warning Resolution field "@parcel/runtime-js@2.8.3" is incompatible with requested version "@parcel/runtime-js@2.9.3"
```

Maybe we can add `^` to let it know it can install any version above that one

7i7o (approved):
tACK 🚀

PR #324, opened 2024-05-16
fix: misc subscription fixes

nicholaswma:
- Adds two additional routes - manually pay subscriptions + sbscriptions details page (this is temporary for testing to see past txns)
- added testing allowed time in daily, hourly, and 5 minutes 
- misc changes to allow the alarm to work as intended


src/subscriptions/payments.ts
```diff
@@ -182,28 +220,64 @@ const submitTx = async (
...
-    data.nextPaymentDue,
-    data.recurringPaymentFrequency
-  );
+  updatedTxnId: string,
+  manualPayment: boolean = false
+): Promise<SubscriptionData> {
+  let nextPaymentDue: { currentDate: Date; status: "success" | "fail" };
 
-  if (new Date(data.subscriptionEndDate) <= nextPaymentDue) {
+  if (manualPayment) {
```
7i7o, line 239: Why does it matter if the payment was manual to calculate nextPaymentDue?

src/subscriptions/payments.ts
```diff
@@ -17,37 +17,58 @@ import {
...
+  data: SubscriptionData,
+  initialPayment?: boolean
 ): Promise<SubscriptionData | null> {
   const autoAllowance: number = await ExtensionStorage.get(
     "setting_subscription_allowance"
   );
   const activeAddress: string = await ExtensionStorage.get("active_address");
 
-  const subscriptionFee: number = data.subscriptionFeeAmount;
+  // don't charge if auto-renewal is off
```
7i7o: Why? auto-renewal != pay automatically

7i7o (commented):
Left 2 questions in the code

src/subscriptions/payments.ts
nicholaswma: so if a user has manual payments enabled (allowance < fee) and a payment was made 3 days after it was due, the next payment date needs to be when that manual payment was processed

7i7o (approved):
tACK 🚀

PR #329, opened 2024-05-22
refactor: add fallback to switch gateways if primary is offline

pawanpaudel93:
## Summary

This PR adds `https://gw.warp.cc` as the primary gateway and `https://gateway.warp.cc` as a fallback.


7i7o (approved):
utACK 🚀

PR #332, opened 2024-05-23
fix: Ensure transfer quantity is integer and make estimated network fee as USD

pawanpaudel93:
## Summary

This PR ensures that the AR transfer amount is always a integer value. Also, now we calculate the exact fees that is needed for the AR transfer with showing the estimated  network fees as USD.

**Questions:** 
- Was `dummyTarget` used so that we have some AR left after sending all the AR using `Max` ?
- The accepted transfer quantity can be a integer or float depending on the different tokens. When using the max or manual input sometimes the quantity can be a float value and can cause the transfer to fail when the token only accept the integer value. Any thoughts on how we can handle this ?

### DIFF
--- a/src/routes/popup/send/confirm.tsx
+++ b/src/routes/popup/send/confirm.tsx
@@ -182,7 +182,7 @@ export default function Confirm({ tokenID, qty }: Props) {
           quantity: "0"
         });
 
-        const qty = fractionedToBalance(Number(amount), token);
+        const qty = +fractionedToBalance(amount, token, "WARP");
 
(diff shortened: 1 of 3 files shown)


7i7o (approved):
tACK 🚀 
👏 👏 👏

pawanpaudel93:
@7i7o AR and AO token transfer quantity should not be in exponential notation when the quantity is larger. I will make changes for this.

7i7o (approved):
tACK 🚀

PR #334, opened 2024-05-23
fix: Disabled allowance bug

pawanpaudel93:
## Summary

This PR fixes the following issues:

- Eliminates continuous loading in AO transaction popup when allowance is disabled.
- Ensures dispatch transactions succeed even when allowance is disabled.
- Displays correct transaction size in the popup.

## How to Test

Disable allowance and test doing AO transactions and dispatch transactions on any Dapp using this PR branch built ArConnect and the ArConnect from ChromeStore. Also, check the transaction size on the popup.


7i7o (approved):
tACK 🚀

PR #339, opened 2024-05-23
fix: pw expiration for hardware wallets

nicholaswma:
- Small bug where adding a keystone wallet first doesn't trigger an expiration and users are then forced to change their pw after onboarding


7i7o (approved):
utACK 🚀

PR #340, opened 2024-05-23
fix: adds enter on password inputs

nicholaswma:
- Note: I've left the ability to hit enter on auth/signDataItem if there's no pw input, lmk if I should remove it


7i7o:
> * Note: I've left the ability to hit enter on auth/signDataItem if there's no pw input, lmk if I should remove it
No, keep that. That's 👍 

Do we have this same behaviour in signMessage and general Sign popups as well?

7i7o (approved):
LGTM 🚀

PR #341, opened 2024-05-24
fix: Upgrade axios to latest for background workers compatibility

pawanpaudel93:
## Summary

Upgrade axios to latest so that `redstone-api` package can work in the background service workers as the latest axios contains the fetch adapter which is selected automatically if other adapter is not found and fetch adapter makes axios run in the background service workers.

## How to Test
- Paste this code below in `src/background.ts`
```ts
import browser, { type Alarms } from "webextension-polyfill";
import redstone from "redstone-api"

async function redstoneHandler(alarm: Alarms.Alarm) {
  if (alarm.name !== "redstone-price") return
  try {
    const value = await redstone.getPrice("AR")
    console.log("Redstone: ", value);
  } catch (e) {
    console.log("Redstone cannot be used.", e);
  }
}

browser.alarms.onAlarm.addListener(redstoneHandler);
browser.alarms.create("redstone-price", { delayInMinutes: 0.1 })
```
- Go to extensions page of browser and click on `service worker` of ArConnect to see the background worker console. If you don't see the redstone value, just reload the extension using reload button on extensions page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "ar-gql": "^0.0.6",
     "arbundles": "^0.9.5",
     "arweave": "^1.13.0",
-    "axios": "^1.1.3",
+    "axios": "^1.7.2",
     "bip39-web-crypto": "^4.0.1",
     "check-password-strength": "^2.0.7",
     "copy-to-clipboard": "^3.3.2",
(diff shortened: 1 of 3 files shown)


7i7o (approved):
tACK 🚀

PR #342, opened 2024-05-24
refactor: Remove token ticker beside token balance in token listing

pawanpaudel93:
## Summary

This PR removes the token ticker beside token balance from the token listing in `Dashboard`, `View all` and `Currency selection` in Send page.


7i7o (approved):
tACK 🚀

PR #343, opened 2024-05-26
fix: updated theme colors for hardware wallets

nicholaswma:
- Manually updating theme colors for primary and hover states
- Updated navigation, send, and receive screen


7i7o (approved):
tACK 🚀

PR #350, opened 2024-05-29
fix: Keystone scanner wait for wallet creation completion

pawanpaudel93:
## Summary

This PR fixes issue with `Generate Wallet` and `Connect Keystone` both using Web Worker at the same time. The QR scanner waits for web Worker to be available again to display the QR Scanner. The waiting time duration for the QR scanner to display depends upon the time the wallet creation takes.


nicholaswma (approved):
tACK 👍

PR #355, opened 2024-05-30
fix: Initialize storage values with defaults to prevent errors

pawanpaudel93:
## Summary

This PR fixes issue with `View all` page when `ao_tokens` storage value might be `undefined` and cause issues with UI.

## How To Test
- Test it on development branch first to see the error. Clear `ao_tokens`, `ao_tokens_cache` and `ao_tokens_ids` first and goto view all page to see the issue.
- Then do the same for this branch PR by clearing the storage values.


7i7o (approved):
tACK 🚀

PR #361, opened 2024-05-31
fix: added gdpr selected for those that are in non-gdpr regions

nicholaswma:
- based locations off https://www.gdpradvisor.co.uk/gdpr-countries


7i7o (approved):
tACK 🚀

PR #364, opened 2024-06-05
refactor: integrate new cu

pawanpaudel93:
## Summary

This PR updates the default AO CU to `https://gateway.ar`. Also, this PR makes sure `syncAoTokens` is run only once even if a user goes to `View all`, goes back and again goes back to `View all`.

### DIFF
--- a/src/tokens/aoTokens/ao.ts
+++ b/src/tokens/aoTokens/ao.ts
@@ -78,6 +78,7 @@ export function useAoTokens(): [TokenInfoWithBalance[], boolean] {
     []
   );
   const [loading, setLoading] = useState(true);
+  const ao = useAo();
 
   const tokensWithBalances = useMemo(
     () =>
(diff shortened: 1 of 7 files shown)


src/tokens/aoTokens/ao.ts
```diff
@@ -144,7 +145,7 @@ export function useAoTokens(): [TokenInfoWithBalance[], boolean] {
               const balance = Number(
                 await timeoutPromise(
                   (async () => {
-                    const aoToken = await Token(id);
+                    const aoToken = await Token(id, null, ao);
```
pawanpaudel93, line 148: @7i7o Should we not use `Token` here and directly fetch balance as this first fetches the token info and then the balance ? Can we directly fetch the balance and use the `Denomination` saved on the storage ?

7i7o: We could, but when could we update the Denomination if someone changes that?

pawanpaudel93: Yes, that can be the issue if we use the stored one and if Denomination of a token changes. We can leave it as it is.

7i7o:
Closing for this was solved in subsequent PRs

PR #366, opened 2024-06-06
feat: Transaction History

nicholaswma:
- Creates a new section tab for "transactions"
- Combines ao and AR sent/receive notifications (40 total max transactions, but we can make this in a way where more transactions can be loaded)
- Transaction details are viewable


7i7o (approved):
tACK 🚀

7i7o (approved):
tACK 🚀

PR #372, opened 2024-06-07
feat: removed password requirements & bypass password changes

nicholaswma:
- Lifted pw requirements aside from 5 character minimum
- Allows user to bypass password change request


7i7o (approved):
tACK 🚀

PR #373, opened 2024-06-10
refactor: subscription analytics

pawanpaudel93:
## Summary

This PR updates the subscription analytics to track data related to new subscription and subscription payment.


pawanpaudel93:
@7i7o  I added what was in the Jira issue but should we track more data for the subscription payment ?

7i7o:
> @7i7o I added what was in the Jira issue but should we track more data for the subscription payment ?

Yes, can we also track when the payment fails? Maybe adding a `Status` to the tracked data if if succeeds or fails

7i7o (approved):
utACK 🚀

PR #377, opened 2024-06-13
fix: Make sure wallet selector stays on top of currency selector

pawanpaudel93:
## Summary

This PR fixes an issue where the currency selector overlaps the wallet selector.


7i7o (approved):
tACK 🚀

PR #380, opened 2024-06-13
fix: bip39 usage issue

firstfan:
The 'getKeyPairFromMnemonic' function from the 'human-crypto-keys' library (v0.1.4) has a bug that may lead to different results in Node.js compared to the browser. This fix ensures consistent results across environments and helps other developers avoid referencing faulty code.


7i7o (approved):
tACK 🚀

7i7o (approved):

7i7o (approved):

PR #381, opened 2024-06-13
fix: Update old ao token with new

pawanpaudel93:
## Summary

This PR removes the old ao token if found and adds the new ao token if not found. Also, it adds the new ao token to the default ao tokens list.

## How to Test
1. Add the old token if not added (If dry-run failed and unable to add the old token then add the new token and update the processId with the old token processId)
2. Run the dev mode and refresh the extension from settings to trigger `onInstalled` to update old ao token with new ao token.
3. You could also remove all the `ao_tokens` from storage and refresh the extension to see the updated defaultAoTokens added.


7i7o (approved):
tACK! 🚀

PR #385, opened 2024-06-14
fix: Add ao token transfer notice popup & disable send flow

pawanpaudel93:
## Summary

This PR adds the following:
- It adds AO token transfer notice popup & disable the AO token transfer flow.
- It also prevents infinite AR balance fetch in Currency selector. (To check the infinite balance fetch, just open the current selector in the Send page and see the Network tab to see infinite balance requests in the beta/prod Arconnect)


7i7o (approved):
tACK 🚀

PR #387, opened 2024-06-14
ArConnect 1.13.2

pawanpaudel93:
### Fixes
- AO token transfer notice popup & disable AO token transfer flow
- Prevent infinite AR balance fetch in Currency selector


7i7o (approved):
LGTM 🚀

PR #398, opened 2024-06-19
fix: Use correct ao link message url

pawanpaudel93:
## Summary

This PR fixes the usage of correct ao link message url.


7i7o (approved):
tACK 🚀

PR #401, opened 2024-06-21
arc-394:  Adds Token Info Fetching 

nicholaswma:
- Balances on dashboard are directly queried rather than querying for token info first
- Token Info are queried once a week with manual refreshing able to be called in the Token Settings page
- Balance will call token info if users are on Send page for accuracy 
- signDataItem auth screen will only query for Token Information if it's a `Transfer` (tighter params can be added)

### DIFF
--- a/src/components/dashboard/subsettings/TokenSettings.tsx
+++ b/src/components/dashboard/subsettings/TokenSettings.tsx
@@ -1,5 +1,6 @@
 import {
   ButtonV2,
+  Loading,
   SelectV2,
   Spacer,
   Text,
@@ -13,7 +14,7 @@ import { ExtensionStorage } from "~utils/storage";
 import { AnimatePresence } from "framer-motion";
 import { TrashIcon } from "@iconicicons/react";
 import { removeToken } from "~tokens";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import CustomGatewayWarning from "~components/auth/CustomGatewayWarning";
 import { CopyButton } from "./WalletSettings";
 import browser from "webextension-polyfill";
@@ -24,6 +25,7 @@ import { formatAddress } from "~utils/format";
 import { ResetButton } from "../Reset";
 import { RefreshCcw01 } from "@untitled-ui/icons-react";
 import { defaultAoTokens, type TokenInfo } from "~tokens/aoTokens/ao";
+import TokenLoading from "~components/popup/asset/Loading";
 
 export default function TokenSettings({ id }: Props) {
   // tokens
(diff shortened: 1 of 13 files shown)


src/components/dashboard/subsettings/TokenSettings.tsx
```diff
@@ -73,35 +77,78 @@ export default function TokenSettings({ id }: Props) {
     });
   }
 
+  const refreshToken = async () => {
+    const defaultToken = defaultAoTokens.find((t) => t.processId === token.id);
```
pawanpaudel93, line 86: Why not refresh token info for all ao tokens present ?

src/components/dashboard/subsettings/TokenSettings.tsx
```diff
@@ -120,25 +167,59 @@ export default function TokenSettings({ id }: Props) {
...
           </AnimatePresence>
         </div>
       )}
-      <ButtonV2 fullWidth onClick={() => removeToken(id)}>
-        <TrashIcon style={{ marginRight: "5px" }} />
-        {browser.i18n.getMessage("remove_token")}
-      </ButtonV2>
+      <ButtonWrapper>
+        {isAoToken && (
+          <ButtonV2
```
pawanpaudel93, line 187: Adding loading spinner to the Refresh Token will help to know if refresh is in progress.

src/tokens/aoTokens/ao.ts
```diff
@@ -153,31 +166,47 @@ export function useAoTokens(): [TokenInfoWithBalance[], boolean] {
...
-                      process: AO_NATIVE_TOKEN_BALANCE_MIRROR,
-                      tags: [{ name: "Action", value: "Balance" }]
-                    });
-                    const balance = res.Messages[0].Data;
-                    if (balance) {
-                      return new Quantity(
-                        BigInt(balance),
-                        BigInt(12)
-                      ).toString();
+              const balance = Number(
```
pawanpaudel93: balance should not be converted to Number as precision is lost with it. Please, keep it as string.

src/components/dashboard/subsettings/TokenSettings.tsx
nicholaswma: I'm not opposed to that - I just did it this way since it's on an individual token page

pawanpaudel93: @nicholaswma Yes, its on an individual token page but its refreshing for only tokens that are the `defaultTokens`. For `!defaultToken` the refresh is not handled. I mean can't we remove the default token check to refresh info for every token individually ?

nicholaswma: good catch - I meant it to be the other way around, my intention was for default tokens to not have to be refreshed (this is essentially to not querying the ao token process)

PR #402, opened 2024-06-21
fix: Fetch token info to show readable ticker and amount

pawanpaudel93:
## Summary

This PR fetches token info for tokens not found in storage to ensure notifications and transaction history display the human readable ticker and amount.

## Optimization
Currently, tokens not found in storage are fetched and cached only for the duration of a session, meaning they are re-fetched each time a new popup is triggered. To enhance efficiency, should we consider caching this information in storage, or do you have any other suggestions for improving this process?


7i7o (approved):
tACK 🚀

PR #404, opened 2024-06-21
fix: added copy ability on wallet switcher

nicholaswma:
Adds copy icon to wallet switcher

image


7i7o (approved):
This is a MAJOR update 🙌 😄 

tACK 🚀

PR #405, opened 2024-06-21
fix: better error messages on import

nicholaswma:
Removes error toasts on import wallet, adds error messages for various states(pw, no pw, no kf, etc)

image

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -1997,5 +1997,21 @@
   "recommended":  {
     "message": "Recommended",
     "description": "required"
+  },
+  "add_wallet_error_message": {
+    "message": "Please enter a recovery phrase or key file and enter the password to your current wallet before continuing.",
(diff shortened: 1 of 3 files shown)


assets/_locales/en/messages.json
```diff
@@ -1997,5 +1997,21 @@
   "recommended":  {
     "message": "Recommended",
     "description": "required"
+  },
+  "add_wallet_error_message": {
+    "message": "Please enter a recovery phrase or key file and enter the password to your current wallet before continuing.",
+    "description": "Error message if recovery section is left blank"
+  },
+  "empty_password_error_message": {
+    "message": "Please enter the password to your current wallet before continuing. Not the password to the wallet you are importing.",
```
7i7o: Not a fan of this copy. Maybe we could signal somewhere that the password is *Arconnect*-bound instead of *wallet*-bound

7i7o (approved):
tACK 🚀

PR #406, opened 2024-06-24
refactor: Show network connectivity error tooltip

pawanpaudel93:
## Summary

This PR adds the network connectivity tooltip for AO tokens when the balance fetch is failed.


src/components/popup/Token.tsx
```diff
@@ -228,6 +232,17 @@ const DegradedMessage: React.ReactNode = (
   </div>
 );
 
+const NetworkErrorMessage: React.ReactNode = (
+  <div style={{ textAlign: "center" }}>
+    <div style={{ fontSize: "14px" }}>Network issue detected.</div>
+    <div style={{ fontSize: "12px", color: "#a3a3a3" }}>
+      Your ISP or VPN is blocking access to AO.
```
7i7o: Let's i18n these two messages

7i7o (changes requested):

src/components/popup/Token.tsx
pawanpaudel93: Sure, I will do that.

pawanpaudel93:
@7i7o since we cannot get the same fetch error status as we see in the browser, we don't know if it was failed due to ISP/VPN blocking ao. I did add a common fetch error check that occurs on if fetch is failed but do you have idea on how we can make it better ?

7i7o:
> @7i7o since we cannot get the same fetch error status as we see in the browser, we don't know if it was failed due to ISP/VPN blocking ao. I did add a common fetch error check that occurs on if fetch is failed but do you have idea on how we can make it better ?

Actually those are the best error messages we have from the users that experienced the problem

7i7o (approved):
tACK 👍

PR #407, opened 2024-06-25
feat: Auto import AO tokens in dashboard

pawanpaudel93:
## Summary

This PR adds functionality to auto import AO tokens to dashboard. If a AO token is removed from dashboard using `View All` or `Token Settings` then the token is not auto imported again.

`Note:` Also, now `arweave-search.goldsky.com` graphql endpoint is used so the graphql querying is faster than before.


7i7o (approved):
tACK 👍

nicholaswma:
@pawanpaudel93 can you clear out these conflicts? thanks

7i7o (approved):
tACK 👍

PR #412, opened 2024-07-01
refactor: integrate Ar.io ao process

pawanpaudel93:
## Summary

This PR refactors the codebase to use the new Ar.io AO process instead of the api.


src/lib/arns.ts
```diff
@@ -1,41 +1,144 @@
...
+      })
+    )
+  );
+
+  if (results.length === 0) {
+    return [];
+  }
+
+  return Array.from(new Set(results.filter((result) => result)));
+};
```
7i7o, line 126: We need to find another way to do this (maybe collaborating with Ar.IO), or consider removing it from the contacts' info

pawanpaudel93: It has been commented out for now.

7i7o (approved):
tACK 👍

PR #416, opened 2024-07-03
feat: new settings UI

pawanpaudel93:
## Summary

This PR adds the new quick settings and makes modifications on the All settings page to show the General/Advanced settings.


src/routes/popup/settings/apps/[url]/index.tsx
Comment (7i7o) [inline src/routes/popup/settings/apps/[url]/index.tsx:1]: When the app has a different port, for example localhost:5173, the colon is written as "%3A" and it breaks the route resulting in the setting not being shown

7i7o (changes requested):
Only one minor fix for the apps settings routes

7i7o (approved):
tACK 👍

PR #418, opened 2024-07-03
fix: Endless spinner issue when loading AR balance on dashboard

pawanpaudel93:
## Summary

This PR fixes the endless spinner issue when loading AR balance on dashboard by comparing the float balance value to historical balance as the historical balances are float value of balances.

## How to Test
- Test the balances of the current wallets.
- Test the balance of the wallet where the issue was occurring.


7i7o (approved):
tACK 👍

PR #421, opened 2024-07-05
fix: Show correct fiat amount for AR in transactions history

pawanpaudel93:
## Summary

This PR fixes the transactions history to show the correct fiat amount for AR transactions.


7i7o (approved):
tACK 👍

PR #425, opened 2024-07-09
fix: adds delete and get subscription route

nicholaswma:


7i7o (approved):
tACK 👍 

Usage issues:
- devs can read *all* subscriptions if they have the permissions
- deleteSubscription actually *removes* the subscription instead of *canceling* from the list of subscriptions (user can no longer see their past payments or previous subscription)
- deleteSubscription doesn't show popup neither needs any user confirmation
- devs can delete *any* subscription (without user confirmation) if they know the `targetId`

Previous subscription rough edges:
- If user cancels a subscription, they can no longer subscribe to the same address: Console shows `Account is already subscribed`. Users need to remove subscription (loosing all subscription payments history) to be able re-subscribe.

nicholaswma:
Closing for now - can reopen if necessary

PR #433, opened 2024-07-10
Show import wallet key length error

pawanpaudel93:
## Summary

This PR adds the error modal shown when importing non-supported wallets other than 4096 bits wallet. Also, it adds a fix to not show the two unlock ArConnect popup after trying to open ArConnect when a new wallet is generated or imported for the first time occurring due to `checkBits` function.


7i7o (approved):
tACK 👍

PR #436, opened 2024-07-11
Show wallet creation error when key length mismatched

pawanpaudel93:
## Summary

This PR shows the wallet generation error modal when actual key length doesn't equals to expected key length with retry action key to retry wallet generation. Also, it fixes the password matching logic.

## How To Test

- Maybe modify this https://github.com/arconnectio/ArConnect/blob/dc1d616e549303b3616c68ef88fc4e07984764af/src/wallets/generator.ts#L27 to choose the random keys length `modulusLength: [4096, 2048][Math.floor(Math.random() * 2)]` to test this PR so that the error modal shows.


7i7o (changes requested):
Can we change the messaging?
- Import/Add wallet from JWK: Error message 👍 
- Creating/Importing wallet from a seedphrase:
    - automatically retry
    - maybe show a message that it's taking longer than usual

Whenever we create the key (from a seedphrase), it should always be valid.
Whenever a key comes from the user, then we should show an error for bad keys.

7i7o (approved):
tACK 👍

PR #439, opened 2024-07-12
Arc 420/permissions screen 1

nicholaswma:
**Allowances:**
- Disabled + Always Allow: Allowance is disabled - limit is infinite
- Enabled + Always Allow: Allowance is set whatever is in the input with a minimum of 0.001
- Disabled + Always Ask: Allowance is enabled, limit is set to 0
- Enabled + Always Ask: Allowance is enabled, limit is set to 0

Always Ask is determined by: `const alwaysAsk = allowance.enabled && allowance.limit.eq(BigNumber("0"));`

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -953,6 +953,10 @@
     "message": "Always ask permission",
     "description": "Always allow button"
   },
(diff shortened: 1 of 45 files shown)


7i7o (approved):
tACK 👍

PR #453, opened 2024-07-31
refactor: Add retries & toast error when quote fetch failed

pawanpaudel93:
## Summary

This PR adds the following:
- Retries to fetching Transak quote
- Show error toast when Transak quote fetch failed like due to:
  - api error (server not responding)
  - client error (showing error on minimum amount required or payment not supported)


7i7o (approved):
tACK 👍

PR #454, opened 2024-07-31
feat: adds token balances endpoint

nicholaswma:
@7i7o **This only returns AO tokens and balances - do we want to return balance for AR?**

TODO: we'll need to update types and docs


7i7o (approved):
tACK 👍

src/api/modules/user_tokens/index.ts
```diff
@@ -0,0 +1,11 @@
+import type { PermissionType } from "~applications/permissions";
+import type { ModuleProperties } from "~api/module";
+
+const permissions: PermissionType[] = ["ACCESS_BALANCES"];
```
pawanpaudel93: Should this permission still be called `ACCESS_BALANCES` when the api is renamed ? What about `ACCESS_TOKENS` ?

pawanpaudel93: @7i7o Thoughts?

7i7o: I agree with @pawanpaudel93 that `ACCESS_TOKENS` is a better name for the added permission

pawanpaudel93 (approved):
tACK 🚀

Just a suggestion for the permission text: 
> Allows access to view tokens and their balances in your wallet.

7i7o (approved):
tACK 👍

PR #463, opened 2024-09-05
fix: added arconnect tag to ao transfers

nicholaswma:


src/tokens/aoTokens/ao.ts
```diff
@@ -422,7 +423,9 @@ export const sendAoTransfer = async (
           name: "Recipient",
           value: recipient
         },
-        { name: "Quantity", value: amount }
+        { name: "Quantity", value: amount },
+        { name: "Client", value: "ArConnect" },
+        { name: "Client-Version", value: browser.runtime.getManifest().version }
```
7i7o, line 428:
```suggestion
        { name: "Signing-Client", value: "ArConnect" },
        { name: "Signing-Client-Version", value: browser.runtime.getManifest().version }
```

Let's use this Tag names instead. Ref: https://viewblock.io/arweave/tx/yqzbBl88nuyZaVHqvVFTDchKYGdruvGd9g4ciVRnwMo

src/tokens/aoTokens/ao.ts
```diff
@@ -524,7 +527,9 @@ export const sendAoTransferKeystone = async (
...
```
7i7o, line 532:
```suggestion
        { name: "Signing-Client", value: "ArConnect" },
        { name: "Signing-Client-Version", value: browser.runtime.getManifest().version }
```

Same as previous comment

7i7o (changes requested):
Just a small Tag name change 😊

src/tokens/aoTokens/ao.ts
nicholaswma: @7i7o just to confirm AR transfers have used Client and Client-Version and ao transfers will use Signing-Client and Signing-Client-Version?

7i7o: Yes, I was thinking about distinguishing both because one would be for an actual AR transaction, while the other could be used to sign AO messages/DataItems
Wdyt?

7i7o (approved):
utACK 👍

PR #464, opened 2024-09-09
fix: Throw error on ao token negative & zero amount quantity

pawanpaudel93:
## Summary

This PR fixes the ArConnect popup crash by throwing error when AO token quantity is negative, non-number or zero.

## How to Test
1. Goto [AO Link](https://www.ao.link/#/entity/agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA?tab=write), scroll down and connect the wallet and send the following message pasting on the Query.

  ```json
  {
    "process": "agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA",
    "data": "",
    "tags": [
      {
        "name": "Action",
        "value": "Transfer"
      },
        {
        "name": "Recipient",
        "value": "7waR8v4STuwPnTck1zFVkQqJh5K9q9Zik4Y5-5dV7nk"
      },
        {
        "name": "Quantity",
        "value": "-1"
      }
    ]
  }
  ```
  2. Repeat 1. for both this PR and development branch.

### DIFF
--- a/src/api/modules/sign_data_item/sign_data_item.background.ts
+++ b/src/api/modules/sign_data_item/sign_data_item.background.ts
@@ -46,18 +46,22 @@ const background: ModuleFunction<number[]> = async (
     )
   ) {
     try {
-      const tags = dataItem?.tags || [];
-      const quantityTag = tags.find((tag) => tag.name === "Quantity");
+      const quantityTag = dataItem.tags?.find((tag) => tag.name === "Quantity");
       if (quantityTag) {
-        const quantity = BigNumber(quantityTag.value).toFixed(
-          0,
-          BigNumber.ROUND_FLOOR
-        );
-        if (!isNaN(+quantity)) {
-          quantityTag.value = quantity;
+        const quantityBigNum = BigNumber(quantityTag.value);
+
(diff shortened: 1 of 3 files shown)


7i7o (approved):
tACK 👍

femaury:
If I may, I don't believe it is ArConnect's place to decide what is a valid Transfer Quantity or not.
This PR prevents anyone who has developed processes which use a Action: Transfer message with non positive number values as Quantity to be compatible with ArConnect...

This isn't how the crash should have been handled imo.

7i7o:
Hey @femaury, thanks so much for your feedback!

For this specific issue, we focused on AO token transfers, and our goal is to ensure compliance with the [token specs](https://cookbook_ao.arweave.net/references/token.html#token-processes). As part of that, Transfer Actions are expected to include "1 or more units" in the Quantity Tag, meaning that non-positive quantities wouldn't align with the spec.

I understand there are different approaches and processes that could use a different spec, which could lead to a more philosophical discussion on the definition of a token. However, for now, we kept our scope narrow and didn't delve into that aspect.

Thanks again for your input—really appreciate it!

PR #482, opened 2024-10-01
fix: Prevent popup freeze by sending keep-alive alarm to background

pawanpaudel93:
## Summary

This PR sends keep-alive alarm to background every 20s to keep service worker alive so that the popup doesn't freeze when kept open for long time.

## How to Test
- Test both this PR (test with build not dev) and the production/beta extension with ao.link and send an AO message with write tab for a process [HERE](https://www.ao.link/#/entity/txvY5I2PDnBCW6vHpVxPN3m-SKP-ylJu0w7D_priUhs?tab=write). Keep the popup for long time (> 10 minutes or maybe longer) then approve or cancel the popup to see if it works or not.


7i7o (approved):
tACK 👍 Was kept alive even after 5 hrs

PR #484, opened 2024-10-04
Fix some UI changes.

### LINKED ISSUE #474: Constant resizing due to appearing/disappearing scrollbar
When navigating across the different views, the scrollbars appear/disappear depending on the content, causing a minor but annoying resize.

### LINKED ISSUE #475: Plenty of dead space around some buttons / clickable items
- Launcher buttons (bottom bar) - Unless you click almost dead center on the button icon/text, the click won't register.
- "Explore" view items - Only the logo and the "external link" icon on the right are clickable, but not the item itself.

Not having a hover effect (other then the "pointer" cursor) on some of the above doesn't help either.

### LINKED ISSUE #476: Inconsistent "Explore" view layout
This view will hide the bottom launcher/navigation and instead just show a "back" arrow button on the top-left corner, as opposed to all the other views.

### LINKED ISSUE #477: Inconsistent popup underlay
- The wallet dropdown has an underlay that is totally transparent, so the other elements on the view are totally visible but not interactable.
- The "three dots" dropdown doesn't have an underlay, so the other elements on the view are totally visible and interactable.

Always showing an underlay that fades/blurs the elements behind might help reduce visual noise and prevent missclicks when users only want to close the popup.

### LINKED ISSUE #479: When clicking "Max" in the "Send" view with a wallet with no AR, "NaN" is shown
image

Danziger:
- Close https://github.com/arconnectio/ArConnect/issues/474.
- Close https://github.com/arconnectio/ArConnect/issues/475.
- Close https://github.com/arconnectio/ArConnect/issues/476.
- Close https://github.com/arconnectio/ArConnect/issues/477.
- Close https://github.com/arconnectio/ArConnect/issues/479.
- Tone down screen animations (replaced slide in/out with fade in/out).
- Replace some clickable `<div>` with `<button>`.

Wallet selector with underlay:

Menu with underlay:

"Send" recipient selector drawer:

"Send" currency selector drawer (label to be updated):

"Buy" currency selector drawer (label to be updated):

"Buy" payment method selector (label to be updated):


src/components/SliderMenu.tsx
```diff
@@ -1,48 +1,100 @@
...
   flex-direction: column;
-  min-height: 100%;
-  padding-bottom: 62px;
+  padding-bottom: 16px;
   width: 100%;
+  z-index: 1000;
+  overflow: scroll;
   background-color: ${(props) =>
     props.displayTheme === "light" ? "#ffffff" : "#191919"};
+  border-radius: 10px 10px 0 0;
```
7i7o, line 81: The white background mixed with the white in the CloseLayer makes this really hard to see in Light mode
Wdyt about maybe adding a subtle shadow or border to make it easier to see the slider covering the CloseLater/background items?

src/components/popup/home/Tabs.tsx
```diff
@@ -70,15 +70,17 @@ const TabsWrapper = styled.div`
...
-const StyledTab = styled.div<{ active?: boolean; tabId: number }>`
+const StyledTab = styled.button<{ active?: boolean; tabId: number }>`
   position: relative;
   height: 25px;
   font-weight: 500;
   font-size: 18px;
   line-height: 25px;
-  color: ${(props) =>
-    props.active ? props.theme.primaryText : props.theme.secondaryTextv2};
+  color: ${(props) => (props.active ? "white" : props.theme.secondaryTextv2)};
```
7i7o:
```suggestion
  color: ${(props) => (props.active ? props.theme.primaryText : props.theme.secondaryTextv2)};
```

Because in Light Mode this ends up being white on white

7i7o (changes requested):
Left comments for minor UI issues in Light Mode

src/components/SliderMenu.tsx
nicholaswma: moved the slider to go higher which shows less of the components behind it, we can add a border if it's still too distracting

nicholaswma (approved):
tACK

src/components/SliderMenu.tsx
```diff
@@ -1,48 +1,98 @@
...
 const ExitButton = styled(CloseIcon)`
   cursor: pointer;
 `;
 
-const Wrapper = styled.div<{ displayTheme: DisplayTheme }>`
+const Wrapper = styled(motion.div)<{ displayTheme: DisplayTheme }>`
   position: fixed;
+  bottom: 0;
+  left: 0;
+  height: 93%;
```
7i7o: The fixed height doesn't look right when you open Payment Methods or when you don't have many recent txs or contacts and you open recipient
Maybe keeping `height: auto` and increasing to `max-height: 93%` instead?

src/components/popup/home/Tabs.tsx
```diff
@@ -79,6 +79,9 @@ const StyledTab = styled.div<{ active?: boolean; tabId: number }>`
   color: ${(props) =>
     props.active ? props.theme.primaryText : props.theme.secondaryTextv2};
```
7i7o:
```suggestion
    props.active ? props.theme.primaryTextv2 : props.theme.secondaryTextv2};
```
`primaryText` is not rendering properly for me, and it would be better to use the new v2 colors anyway

7i7o (changes requested):

src/components/SliderMenu.tsx
nicholaswma: @7i7o hmm - I think it was more so my preference, I'll move it back to auto but I changed it because a low sliding slider didn't feel balanced to me 

image

7i7o: Yeah, I agree that it doesn't look so good in this case, but I guess that's because it also ends in a heading. Maybe we could create a task to put a `<No Contacts>` text/link to add some or something similar so it's not that unbalanced

PR #493, opened 2024-10-13
Support archiving/printing entire websites to Arweave

martonlederer:
This PR adds a virtual printer to the existing printers' list in Chrome. Once selected, the printer archives the PDF version of the website/content on Arweave using Turbo.

Example website uploaded to Arweave:
https://arweave.net/1WeS6vpKYid0jizwsGBCjWUEcTVEVXoj6xNd4UpZIEw

Preview:
Screenshot 2024-10-13 at PM 12 38 14

### DIFF
--- a/src/lib/printer.ts
+++ b/src/lib/printer.ts
@@ -1,4 +1,11 @@
+import { getActiveKeyfile, type DecryptedWallet } from "~wallets";
+import { freeDecryptedWallet } from "~wallets/encryption";
+import { concatGatewayURL } from "~gateways/utils";
+import { findGateway } from "~gateways/wayfinder";
 import browser from "webextension-polyfill";
+import Arweave from "arweave";
+
+const ARCONNECT_PRINTER_ID = "arconnect-permaweb-printer";
 
 /**
  * Tells Chrome about the virtual printer's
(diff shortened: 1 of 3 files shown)


src/lib/printer.ts
```diff
@@ -0,0 +1,172 @@
...
+    await uploadDataToTurbo(dataEntry, "https://turbo.ardrive.io");
+
+    // this has to be one of FAILED, INVALID_DATA, INVALID_TICKET, OK
+    resultCallback("OK");
+
+    // find a gateway to display the result
+    const gateway = await findGateway({});
+
+    // open in new tab
+    await chrome.tabs.create({
```
7i7o, line 156: We should add a little delay before opening the posted dataItem
In one of my tests, the gateway didn't have the data ready and received a "Not found" after printing

7i7o (approved):
tACK 👍

t8:
This idea is insane. Love it.

src/lib/printer.ts
martonlederer: Should I add that in this PR? :)

src/lib/printer.ts
```diff
@@ -0,0 +1,172 @@
...
+    const dataSigner = new ArweaveSigner(decryptedWallet.keyfile);
+    const dataEntry = createData(
+      new Uint8Array(await data.arrayBuffer()),
+      dataSigner,
+      { tags }
+    );
+
+    // sign an upload data
+    await dataEntry.sign(dataSigner);
+    await uploadDataToTurbo(dataEntry, "https://turbo.ardrive.io");
```
pawanpaudel93, line 147: @martonlederer Since Turbo has a limit of ~100KB for free uploads, and additional credits are required for larger files, we currently can't upload files larger than 100KB using Turbo alone if we haven't topup sufficient credits. Should we consider falling back to direct Arweave uploads if Turbo upload fails ? WDYT?

martonlederer: I'm honestly not sure. I think bundled txes should be preferred, but what you're saying makes complete sense.

pawanpaudel93: Yes bundled txes should be preferred, but `Print to Arweave` may not work for most cases, as the data size will likely exceed the 100KB limit and users mayn't have Turbo credits. The only pages I could print were simple pages with minimal content otherwise it failed to upload. (**Note**: I don't have Turbo credits/balance.)

7i7o: I agree @pawanpaudel93, let's fall back to Arweave when Turbo upload fails

pawanpaudel93:
Merging to `arconnectio:feat/print-to-arweave` to push some new changes.

PR #494, opened 2024-10-15
Move CSS reset, font loading and theme switch outside React to prevent flicker

Danziger:
- Add different icons for development (plasmo does that automatically but we were overriding that from JS).
- Move CSS reset and font definitions to a CSS file to remove font loading flicker.
- Unify theme logic into a single provider and persist background color in `localStorage` to remove theme flicker.


src/components/popup/Route.tsx
```diff
@@ -26,10 +26,9 @@ export const Wrapper = styled(motion.div)<{
   expanded?: boolean;
 }>`
   position: relative;
-  width: ${(props) => (props.responsive ? "100%" : "377.5px")};
```
Danziger, line 29: This `.5` seem to fix some rendering issues (underflow) in some cases but cause another issue (overflow) in some other cases. I'd just leave it as `377px` which is the actual value this should have everywhere and fix it later by laying out the components differently.

src/components/popup/Route.tsx
```diff
@@ -26,10 +26,9 @@ export const Wrapper = styled(motion.div)<{
...
+  width: ${(props) => (props.responsive ? "100%" : "377px")};
   min-height: ${(props) => (props.expanded ? "100vh" : "600px")};
   max-height: max-content;
-  background-color: rgb(${(props) => props.theme.background});
```
Danziger, line 32: The background is set on the `body` by the `ThemeProvider` in the components library, so no need to set it again here.

src/popup.tsx
```diff
@@ -67,167 +68,141 @@ export default function Popup() {
...
-                  )}
-                </Route>
-                <NavigationBar />
-              </HistoryProvider>
-            </Router>
-          </Wrapper>
-        </ExpandedViewWrapper>
-      </HardwareWalletTheme>
-    </Provider>
+    <ArConnectThemeProvider>
```
Danziger: `ArConnectThemeProvider` replaces `Provider` + `HardwareWalletTheme` and used the `useTheme` hook internally.

src/routes/popup/transaction/[id].tsx
```diff
@@ -356,7 +372,50 @@ export default function Transaction({ id: rawId, gw, message }: Props) {
                     {browser.i18n.getMessage("transaction_from")}
                   </PropertyName>
                   <PropertyValue>
-                    {formatAddress(transaction.owner.address, 6)}
+                    <div>
```
Comment (Danziger) [inline src/routes/popup/transaction/[id].tsx:375]: Eventually, it would be better to turn this into a bunch of individual components.

In any case, the reason for this change is that before the UI would only prompt users to add the "to" address as contact. After this change, it would prompt for both "to" and "from", and it would also take into account if an address is yours and display "(you)" next to it.

src/components/SliderMenu.tsx
```diff
@@ -7,13 +7,15 @@ import styled, { useTheme } from "styled-components";
...
   title: string;
+  hasNav?: boolean;
   isOpen: boolean;
   onClose?: () => void;
   children?: React.ReactNode;
 }
 
 export default function SliderMenu({
   title,
+  hasNav,
```
Danziger, line 18: In the Send view, the last option inside the menus was covered by the bottom navigation.

src/components/hardware/HardwareWalletTheme.tsx
```diff
@@ -1,38 +1,81 @@
...
+  return theme;
+}
+
+export function ArConnectThemeProvider({ children }: PropsWithChildren<{}>) {
   const hardwareApi = useHardwareApi();
+  const theme = useTheme();
+  const themeModifier = hardwareApi ? hardwareThemeModifier : noThemeModifier;
+
+  useEffect(() => {
+    const reducedMotionPreference = window.matchMedia(
```
Danziger, line 34: Added this here. It could eventually be added in settings too, just like the system theme can be overridden.

src/components/popup/HeadV2.tsx
```diff
@@ -110,37 +108,40 @@ export default function HeadV2({
...
     >
-      <BackButton
-        onClick={async () => {
-          if (back) await back();
-          else goBack();
-        }}
-      >
-        <BackButtonIcon />
-      </BackButton>
+      {showBack ? (
```
Danziger, line 111: The conditional was missing here, so that option did nothing.

src/components/popup/HeadV2.tsx
```diff
@@ -110,37 +108,40 @@ export default function HeadV2({
...
-              <HardwareWalletIcon
-                icon={keystoneLogo}
-                color="#2161FF"
-                {...hwIconAnimateProps}
-              />
-            )}
-          </AnimatePresence>
-        </ButtonAvatar>
-      </AvatarButton>
+      {showOptions ? (
```
Danziger, line 124: This other one was probably implemented incorrectly and `allowOpen` (now removed) and `showOptions` were redundant but not implemented properly. No, the lock screen should only show the title at the top, no back button or wallet selector.

src/components/popup/HistoryProvider.tsx
```diff
@@ -27,18 +26,6 @@ export default function HistoryProvider({ children }: PropsWithChildren<{}>) {
     history.back();
   };
 
-  // redirect to unlock if decryiption
-  // key is not available
-  useEffect(() => {
```
Danziger, line 32: Logic moved to `useSetUp()` in `ArConnect/src/wallets/index.ts`.

src/components/popup/Route.tsx
```diff
@@ -14,32 +14,18 @@ const Route: typeof BaseRoute = ({ path, component, children }) => {
     ? children(params)
     : children;
 
-  return (
-    <AnimatePresence initial={false}>
```
Danziger, line 18: Added a single `AnimatePresence` in `popup.tsx`.

src/components/popup/Route.tsx
```diff
@@ -14,32 +14,18 @@ const Route: typeof BaseRoute = ({ path, component, children }) => {
...
-}>`
+const PageWrapper = styled(motion.main)`
   position: relative;
-  width: ${(props) => (props.responsive ? "100%" : "377.5px")};
-  min-height: ${(props) => (props.expanded ? "100vh" : "600px")};
-  max-height: max-content;
-  background-color: rgb(${(props) => props.theme.background});
-`;
-
-const PageWrapper = styled(Wrapper)`
```
Danziger, line 35: There were some unnecessary containers (HTML) that is now gone, so the HTML structure should look much simpler:

    body > #__plasmo > main#Page

src/components/popup/WalletHeader.tsx
```diff
@@ -242,18 +234,23 @@ export default function WalletHeader() {
         }
       }
     ];
-    if (!isExpanded && items.length === 4) {
+
+    if (location.pathname === "/popup.html") {
```
Danziger, line 238: The full screen version is now loaded from a different HTML file: `fullscreen.html`, so all these checks are now much simpler and the styling changes require no JS.

src/components/popup/home/Balance.tsx
```diff
@@ -111,18 +111,9 @@ export default function Balance() {
...
-  // router push
-  const [push] = useHistory();
-
   // display theme
   const theme = useTheme();
 
-  // lock wallet and terminate session
-  async function lockWallet() {
-    await removeDecryptionKey();
-    push("/unlock");
```
Danziger, line 123: No need for a route change anymore. Removing the decryption key makes the root re-render automatically, showing the Unlock screen.

src/popup.tsx
```diff
@@ -49,187 +46,158 @@ import ContactSettings from "~routes/popup/settings/contacts/[address]";
...
+import { ArConnectThemeProvider } from "~components/hardware/HardwareWalletTheme";
+import { AnimatePresence } from "framer-motion";
 
 export default function Popup() {
-  const theme = useTheme();
-  const [expanded, setExpanded] = useState(false);
-
-  // init popup
-  useSetUp();
+  const initialScreenType = useSetUp();
```
Danziger, line 53: The route guarding logic now lives here and doesn't rely on routes to work. That applies to the Unlock screen and an "undefined" state when the app first loads. This will also include a state and UI for the "creating wallet" state of the embedded wallet.

src/popup.tsx
```diff
@@ -49,187 +46,158 @@ import ContactSettings from "~routes/popup/settings/contacts/[address]";
...
-const HideScrollbar = createGlobalStyle<{ expanded?: boolean }>`
-  * {
-    scrollbar-width: none;
+  let content: React.ReactElement = null;
 
-    &::-webkit-scrollbar {
-      display: none
-    }
+  if (initialScreenType === "cover") {
+    content = <Page />;
```
Danziger, line 62: Empty page. Does nothing but wait.

src/popup.tsx
```diff
@@ -49,187 +46,158 @@ import ContactSettings from "~routes/popup/settings/contacts/[address]";
...
+    content = (
+      <Page>
+        <Unlock />
+      </Page>
+    );
+  } else if (initialScreenType === "generating") {
+    // This can only happen in the embedded wallet:
+    content = (
+      <Page>
+        <p>Generating Wallet...</p>
```
Danziger, line 73: Placeholder for the embedded wallet. This does nothing right now, but I'll update it in the other PR once this one is merged.

7i7o (approved):
tACK 👍

PR #495, opened 2024-10-15
fix: hide fiat value for AO tokens

pawanpaudel93:
## Summary

This PR hides the fiat value for AO tokens to prevent displaying incorrect values due to symbol overlap with other coins/tokens like "AR."


7i7o (approved):
tACK 👍

PR #496, opened 2024-10-15
Add `tokenBalance` api endpoint

7i7o:
## Summary

This PR adds a new function to the api called `tokenBalance` that requires the `"ACCESS_TOKENS"` permissions

You can fetch the balance of a token by providing the token's id:

```typescript
const tokenBalance = await window.arweaveWallet.tokenBalance("<AO_TOKEN_ID>");
```

It should return the token's balance in a `string` in the same format as `userTokens` returns the balances for each token

## Tasks:
- [x] Add `tokenBalance` api
- [x] Add docs https://github.com/arconnectio/docs/pull/7
- [x] Update arconnect types https://github.com/arconnectio/types/pull/5

### DIFF
--- a/src/tokens/aoTokens/ao.ts
+++ b/src/tokens/aoTokens/ao.ts
@@ -16,6 +16,8 @@ import {
 import type { Alarms } from "webextension-polyfill";
 import type { KeystoneSigner } from "~wallets/hardware/keystone";
 import browser from "webextension-polyfill";
+import { fetchTokenByProcessId } from "~lib/transactions";
+import { tokenTypeRegistry } from "~tokens/token";
 
 export type AoInstance = ReturnType<typeof connect>;
 
@@ -223,11 +225,7 @@ export async function getAoTokenBalance(
 ): Promise<Quantity> {
   const aoTokens = (await ExtensionStorage.get<TokenInfo[]>("ao_tokens")) || [];
 
-  const aoToken = aoTokens.find((token) => token.processId === process);
-
-  if (!aoToken) {
-    throw new Error(`Token not found for address '${address}'.`);
-  }
+  let aoToken = aoTokens.find((token) => token.processId === process);
 
   const res = await dryrun({
     Id,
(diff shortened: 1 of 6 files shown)


src/tokens/aoTokens/ao.ts
```diff
@@ -225,6 +225,10 @@ export async function getAoTokenBalance(
 
   const aoToken = aoTokens.find((token) => token.processId === process);
 
+  if (!aoToken) {
+    throw new Error(`Token not found for address '${address}'.`);
```
7i7o: I think this could lead to confusion when a user might remove a token from the cached token list and when the dev asks for the balance it would throw an error, even if the user has balance on that token

pawanpaudel93: @7i7o I think we are using the cached token list for both `userTokens` and `tokenBalance`. In case of `tokenBalance`, if `aoToken` is undefined should we fetch the token info and then fetch and return the balance instead of throwing the error ?

7i7o: I think we should just try to fetch the balance and just raise the error if it throws 🤷 
I wouldn't add the extra Info check, because that info isn't returned, we're adding an extra delay and the dev is passing the actual process Id. If they are not using one of the user's cached tokens, they must check the token process id themselves

pawanpaudel93: So if `aoToken` is `undefined` should we just return the value we get from the process because for the tokens that are present on the cache we do the following to return the balance ?

```typescript
return new Quantity(BigInt(balance), BigInt(aoToken.Denomination));
```

7i7o: Yes, if `aoToken` is `undefined` we should dryRun and return according to the dryRun result:
- dryRun with a balance msg:  `new Quantity(BigInt(balance), BigInt(aoToken.Denomination));`
- dryRun with no messages (means no balance handler): `throw Error`
- dryRun with errors (means maybe the process was not found): `throw Error`

7i7o: No. You are right. We could fetch `Info` to be consistent 👍
Maybe fetch `Info` only after the balance returns a `>0` amount. Wdyt?

pawanpaudel93 (approved):
tACK

nicholaswma (approved):

Issue #472, opened 2024-09-24
"Add $AR to get started" shown even on wallets with balance


Danziger:
Steps to reproduce (from Eric):
- Clear all assets and tokens.
- Click the Home button at the bottom.
- Switch to a wallet with 0 AR.
  - ✅ "Add $AR to get started" is shown.
- Switch to a wallet with 4 AR.
  - ❌ "Add $AR to get started" is still shown, but it shouldn't.
- Click the Send button at the bottom and click the Home button again.
  - ✅ Assets, Collectibles, Transactions section are now correctly displayed instead of "Add $AR to get started".
- Switch to a wallet with 0 AR.
  - ❌ Assets, Collectibles, Transactions section are now incorrectly displayed instead of "Add $AR to get started", but they shouldn't.

PR #498, opened 2024-10-15
fix: Update logic for showing `Add $AR to get started` #472

pawanpaudel93:
## Summary

This PR updates the logic for displaying the message **"Add $AR to get started."**

## How to Test

1. Prepare at least one wallet from each of the following categories:
   - **Wallet with AR:** Contains a balance of AR tokens.
   - **Wallet without AR:** Contains balances of other tokens but no AR.
    - **Wallet without AR and other tokens:** Contains no balances of AR and other tokens.

2. **Testing Wallet Switching and Popup:**
   - Make sure this is done with both scenarios: **all assets** and **zero assets** (delete all assets).
   - Open the ArConnect popup and switch between wallets to observe the UI changes.
   - Open the ArConnect popup for each wallet individually to observe the UI for all wallet types.

3. Ensure all issues from [#472](https://github.com/arconnectio/ArConnect/issues/472) have been resolved.

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -1083,6 +1083,10 @@
     "message": "View all",
     "description": "View all link text"
   },
+  "view_all_assets": {
+    "message": "View all assets",
+    "description": "View all link text"
+  },
   "setting_tokens": {
     "message": "Tokens",
     "description": "Tokens setting title"
--- a/assets/_locales/zh_CN/messages.json
+++ b/assets/_locales/zh_CN/messages.json
@@ -1071,6 +1071,10 @@
       "message": "查看全部",
       "description": "View all link text"
   },
+  "view_all_assets": {
+      "message": "查看所有资产",
+      "description": "View all assets link text"
+  },
   "setting_tokens": {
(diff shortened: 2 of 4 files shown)


7i7o (changes requested):
The second case, a wallet with AO tokens and NO AR tokens, when you remove all AO and Warp tokens it shows the Add $AR and that doesn't let you go to "View All" page to see if you have any tokens that you didn't add

pawanpaudel93:
@7i7o should we change the conditions on when to show `Add $AR to get started` as for the scenario you explained above we might have to check balances for all the tokens that aren't added but discovered in the View all as well ?

7i7o:
> @7i7o should we change the conditions on when to show `Add $AR to get started` as for the scenario you explained above we might have to check balances for all the tokens that aren't added but discovered in the View all as well ?

We can still show the banner while triggering that fetch in the background, and hide the banner if we find some tokens. Wdyt?
Another option is to change the UI for the "View All Assets" to be accesible even when we haven't found any balance yet

nicholaswma (approved):

PR #517, opened 2024-10-21
arc-664: sign confirmation

nicholaswma:
image

NOTE: Network fee will show 0 if it's under 100kb


7i7o (approved):
tACK 👍

pawanpaudel93:
@nicholaswma tACK 🚀 for Print to Arweave. Should we also update the sign tx popup for dispatch as well ? You could try dispatching the images here: https://arweave-wallet-kit-demo.vercel.app/

image

nicholaswma:
@pawanpaudel93 good idea, I'll create a new ticket for that task so we're not blocked with a build for now.

PR #518, opened 2024-10-22
fix: For window is not defined error in the theme preference

Danziger:


pawanpaudel93:
@Danziger Now the service worker is working fine. Still, there is another issue I noticed: Now if you use the injected api, all the popups are either full white or black based on the theme. You could just try to connect wallet to any of the dapps to reproduce the issue.

Danziger:
@pawanpaudel93 It should be fine now, just tested it on Bazar.

7i7o (approved):
tACK 👍

pawanpaudel93 (approved):
tACK 🚀

PR #519, opened 2024-10-22
fix: generate unique wallet names when adding new wallet

pawanpaudel93:
## Summary

This PR makes sure that the new added wallet name is unique.


7i7o (approved):
tACK 👍

PR #520, opened 2024-10-22
feat: Add print archive transactions to notifications and tx history

pawanpaudel93:
## Summary

This PR adds the `Print to Arweave` transactions to notifications and Tx history.

- Check if you can see the `Print archived` message in TX history.
- Check if you can see the new added message in notifications. Try all options of Notifications from settings: `Enable Arweave and ao Transaction Notifications`, `Enable Arweave Transaction Notifications` and `Enable all Arweave and ao Notifications`.

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -2225,5 +2225,65 @@
   "transak_unavailable": {
     "message": "Transak is unavailable at the moment. Please try again later.",
     "description": "Transak unavailable error message"
+  },
+  "print_archived": {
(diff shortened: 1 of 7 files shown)


nicholaswma (approved):
tACK - this is great 🚀

I think in the future we can disable the data on the transaction page since large PrintArchive will show long raw pdf's.

7i7o (approved):
tACK 👍 

Agree with @nicholaswma on not displaying the Data and just providing a link since PDFs will be binary and also don't render properly

PR #525, opened 2024-10-23
fix: disable print archive data & show download data icon

pawanpaudel93:
## Summary

This PR disables loading data for the print arweave transactions showing the Data download icon only when opening the transaction.


nicholaswma (approved):
tACK 🚀

PR #530, opened 2024-10-29
 fix: Use timeout retries for AR balance fetch using different gateways

pawanpaudel93:
## Summary

This PR improves the AR balance fetching process by implementing timeout retries using different gateways.


7i7o (approved):
tACK 👍

PR #535, opened 2024-11-05
fix: resolve token & contact routes black screen

pawanpaudel93:
## Summary

This PR fixes the `Import Token` black screen issue.

### DIFF
--- a/src/components/popup/Route.tsx
+++ b/src/components/popup/Route.tsx
@@ -8,13 +8,15 @@ import styled from "styled-components";
  */
 const Route: typeof BaseRoute = ({ path, component, children }) => {
   const [matches, params] = useRoute(path);
-  const routeContent = component
-    ? createElement(component, { params })
-    : typeof children === "function"
-    ? children(params)
-    : children;
+  const routeContent =
+    matches &&
+    (component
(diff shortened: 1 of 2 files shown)


pawanpaudel93:
@7i7o @nicholaswma This seems to be the issue of matching both paths `/quick-settings/tokens/new` and `/quick-settings/tokens/:id` for `Import Token (/quick-settings/tokens/new)` and rendering both routes on top of each other.

Possible solutions:
1. As the PR
2. Rename the conflicting routes path so it only matches a single route for a path.

wdyt ?

nicholaswma:
@pawanpaudel93 if possible I think renaming the path would be the better course so we don't encourage this type of pattern, but I'm okay with this solution for the time being

pawanpaudel93:
@nicholaswma I used [Switch](https://github.com/molefrog/wouter/tree/v3?tab=readme-ov-file#switch-) now for those overlapping routes so only the first match is returned.

nicholaswma (approved):
tACK - works great! 🚀

7i7o (approved):
tACK 👍 

Also agree with Nick that we should use non-overlapping routes in the future

PR #536, opened 2024-11-08
fix: Use some specific gateways with retry for Print Arweave in tx history

pawanpaudel93:
## Summary

This PR is a fix for showing the `Print to Arweave` transactions in TX History. This might be the temporary fix as most of the gateways are throwing time out error but `goldsky` and `permagate` gateways are working but there might be some delay in showing the txs after its printed.


7i7o (approved):
tACK 👍

PR #539, opened 2024-11-13
fix: filtered out transaction error messages for agent tokens

nicholaswma:


src/notifications/utils.ts
```diff
@@ -69,6 +69,39 @@ query($address: String!) {
...
 `;
 
+export const AF_ERROR_QUERY = `
+query {
+  transactions(
+    first: 10,
+    tags: [
+      {name: "Data-Protocol", values: ["ao"]},
+      {name: "Action", values: ["Transfer-Error"]},
+      {name: "Message-Id", values: ["L-nExyvrfVv5oJQtBnk6VD7-Hgisf-WLC55MMCqQq7A"]},
```
pawanpaudel93: `Message-Id` is static here.

src/lib/transactions.ts
```diff
@@ -87,10 +93,32 @@ const processTransaction = (transaction: GQLEdgeInterface, type: string) => ({
...
 });
 
+export async function checkTransactionError(
+  transaction: GQLEdgeInterface | Transaction
+): Promise<boolean> {
+  if (transaction.node.owner.address !== AGENT_TOKEN_ADDRESS) {
+    return false;
+  }
+
+  return retryWithDelay(async (attempt) => {
```
pawanpaudel93, line 103: I think this might not always return a true or false and sometimes can return an error as well. So, i think we should handle the error cases as well.

Here is one case where there is an error with a gql query so we must handle error cases as well and maybe return false maybe in case of error:

```JSON
{"errors":[{"message":"query timed out"}],"data":null}

```

pawanpaudel93:
@nicholaswma @7i7o  also should we just filter transfer errors for agent tokens only or for all ao tokens ? wdyt ?

7i7o:
> @nicholaswma @7i7o also should we just filter transfer errors for agent tokens only or for all ao tokens ? wdyt ?

Maybe only for the agent at the moment.

src/lib/transactions.ts
nicholaswma: @pawanpaudel93 is this error message under `data.data.transactions.edges`? looking for ways of filtering it out.

nicholaswma: I think if there's an error on the query we just have to leave as if there isn't a "Transfer-Error"

since we're relying on `return data.data.transactions.edges.length > 0;`, it'll return true or false. Or are you referring to something else?

pawanpaudel93: Yes, this is just an example of an error i got for print transactions with timeout error and it is returned from the gql function.
```
{"errors":[{"message":"query timed out"}],"data":null}
```

`data.data.transactions.edges.length > 0` works perfectly fine. I was thinking to add error handling when both retries failed with graphql or we get the timeout error.

pawanpaudel93: I was thinking of modifying to this: Also returning true or false in case of error depends on if we want to show/hide the transaction.

```typescript
export async function checkTransactionError(
  transaction: GQLEdgeInterface | Transaction
): Promise<boolean> {
  if (transaction.node.owner.address !== AGENT_TOKEN_ADDRESS) {
    return false;
  }

  return retryWithDelay(async (attempt) => {
    const data = await gql(
      AF_ERROR_QUERY,
      { messageId: transaction.node.id },
      txHistoryGateways[attempt % txHistoryGateways.length]
    );
    
    if (data?.data === null && (data as any)?.errors?.length > 0) {
      throw new Error(
        (data as any)?.errors?.[0]?.message || "GraphQL Error"
      );
    }
                
    return data.data.transactions.edges.length > 0;
  }, 2).catch(() => false);
}
```

nicholaswma: sounds good

src/lib/transactions.ts
```diff
@@ -87,10 +93,37 @@ const processTransaction = (transaction: GQLEdgeInterface, type: string) => ({
...
+      { messageId: transaction.node.id },
+      txHistoryGateways[attempt % txHistoryGateways.length]
+    );
+
+    if (data?.data === null && (data as any)?.errors?.length > 0) {
+      throw new Error((data as any)?.errors?.[0]?.message || "GraphQL Error");
+    }
+
+    return data.data.transactions.edges.length > 0;
+  }, 2);
```
pawanpaudel93: I think we should attach `.catch(() => false)` to `retryWithDelay` so that in case of gql error even with retries as well we return a boolean not an error.

src/lib/transactions.ts
```diff
@@ -9,10 +9,16 @@ import type { Token } from "~tokens/token";
...
 import { balanceToFractioned, formatFiatBalance } from "~tokens/currency";
+import { AF_ERROR_QUERY } from "~notifications/utils";
+import { gql } from "~gateways/api";
+import { txHistoryGateways } from "~gateways/gateway";
+import { retryWithDelay } from "~utils/retry";
 
 let tokens: TokenInfo[] = null;
 export let tokenInfoMap = new Map<string, TokenInfo | Token>();
 
+const AGENT_TOKEN_ADDRESS = "JwjzLSbwIDIouY0iIMyVEbjMbVMOkL2TTEzdJWwtly8";
```
pawanpaudel93: I think the AGENT token address is `8rbAftv7RaPxFjFk5FGUVAVCSjGQB4JHDcb9P9wCVhQ`
image

Now as the AGENT address here is wrong, the logic for `checkTransactionError` might be wrong as well.

pawanpaudel93 (approved):
tACK 🚀

PR #555, opened 2024-11-29
Isomorphic message utils

Danziger:
- Closes https://communitylabs.atlassian.net/browse/ARC-992.
- Closes https://communitylabs.atlassian.net/browse/ARC-998.
- Closes https://communitylabs.atlassian.net/browse/ARC-1186.
- Includes https://github.com/wanderwallet/Wander/pull/763.

**DONE:**

- [x] Update i18n mock to replace vars.
- [x] Mock `action.setIcon` and `manifest_version`.
- [x] Update `createContextMenus` so that it does't do anything in Embed.
- [x] Update `createAuthPopup` and `connect.background`'s `catch` blocks to re-throw errors so that the `ERR_MSG_NO_WALLETS_ADDED` errors makes it to the client. We might want to whitelist only some specific errors or throw everything in all other API methods (but that implies if we have bugs, their actual details and stack trace will also be logged on the client). - This fixes a bug where calling connect before authenticating would throw a `Cannot destructure property 'authID' of '(intermediate value)' as it is undefined.` error.
- [x] Fix Wander modal not opening automatically when calling API methods without being authenticated (equivalent to opening the Welcome page in BE). 
- [x] Fix app getting stuck in `/auth-request` even after completing the last auth request.
- [x] Fix auth request provider not clearing completed AuthRequests. 
- [x] Fix opening/closing logic so that:
  - New connect requests always open the modal.
  - Other request will no open the modal after the user closes it manually while displaying a pending auth request.
  - Handling all request will no close the modal automatically if it was already open when the request arrived.
- [x] Session storage is cleared on page load to stop some API methods from working before the (re)authentication completes. This also fixes the issue with and old `ExtensionStorage.get("decryption_key")` being used (rather than the new, random one set when the wallets are loaded on app load).
- [x] Fix `sendChunk` so that `sign` works again.
- [x] Add `embedded_connect` and `embedded_disconnect` messages and update button connection indication accordingly. hese are sent from `updateIcon`.
- [x] Update `embedded_balance` to account for hidden balance.
- [x] Make the Sign Out button (under Settings) work. It calls `supabase.auth.signOut` and updates the button text and connection indicator accordingly.
- [x] Update layout defaults so that settings also uses the popup layout.


src/api/background/background-setup.ts
```diff
@@ -36,8 +37,8 @@ export function setupBackgroundService() {
 
   // MESSAGES:
   // Watch for API call and chunk messages:
-  onMessage("api_call", handleApiCallMessage);
-  onMessage("chunk", handleChunkMessage);
+  isomorphicOnMessage("api_call", handleApiCallMessage);
```
Danziger, line 40: Verify handlers cannot be called directly from the SDK (except for the ones that are actually meant to be called from the SDK).

src/components/popup/WalletHeader.tsx
```diff
@@ -526,13 +526,13 @@ const AppAction = styled.div`
   display: flex;
 `;
 
-export const Action = styled(CopyIcon)<{ active?: boolean }>`
+export const Action = styled(CopyIcon)<{ $active?: boolean }>`
```
Danziger, line 529: We should use [transient props](https://styled-components.com/docs/faqs#transient-props-since-51) to avoid propagating undesired props/attributes to the HTML nodes (which will show up as an error on the Console).

src/api/foreground/foreground-setup-ar-protocol-links.ts
```diff
@@ -1,47 +0,0 @@
-import { sendMessage } from "@arconnect/webext-bridge";
-import { isString } from "typed-assert";
-
-// TODO: This is not enough for client-rendered pages (e.g. React apps):
-
-export async function replaceArProtocolLinks() {
```
Danziger, line 6: While this function was injected and called from `contents/ar-protocol.ts`, the `sendMessage` below will never return a result because there was no listener for `"ar_protocol`. Also, it looks like this was replaced by `protocol.handler.ts`, which uses `browser.webNavigation.onBeforeNavigate`.

shim.d.ts
```diff
@@ -14,78 +13,141 @@ declare module "@arconnect/webext-bridge" {
      * `api/foreground/foreground-setup-wallet-sdk.ts` use `postMessage()` to send `arweaveWallet.*` calls that are
      * received in `contents/api.ts`, which then sends them to the background using `sendMessage()`.
      */
-    api_call: ProtocolWithReturn<ApiCall, ApiResponse>;
+    api_call: {
```
Danziger, line 16: This change should most likely be reverted or changed.

src/api/background/handlers/storage/app-config-change/app-config-change.handler.ts
```diff
@@ -1,4 +1,3 @@
-import { sendMessage } from "@arconnect/webext-bridge";
 import type { StorageChange } from "~utils/runtime";
```
Danziger, line 1: We need to test if this handler works with Embed.

src/api/foreground/foreground-setup-embedded-wallet-sdk.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  console.log("setupEmbeddedWalletSDK()", new Date().toISOString());
+
+  /** Init events */
+  const events = mitt<InjectedEvents>();
+
+  // TODO: Can we get the right type here?:
+  const walletAPI = {
+    walletName: IS_EMBEDDED_APP ? "Wander Embedded" : "ArConnect",
+    walletVersion: version,
```
Danziger, line 34: What version should we return here? As the wallet APIs are the same, should this be the same as for the BE?

src/api/modules/connect/connect.background.ts
```diff
@@ -65,11 +65,12 @@ const background: BackgroundModuleFunction<void> = async (
     // add features available after connection
     await updateIcon(true);
     createContextMenus(true);
-  } catch (e: any) {
+  } catch (err) {
     await updateIcon(false);
     createContextMenus(false);
 
-    throw new Error(e);
+    // TODO: Update this in all API methods to get proper stack traces in errors? Maybe only in development?
```
Danziger: Discuss this with @7i7o

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -124,26 +125,27 @@ export function setupWalletSDK(
...
-          if (finalizerResult) {
-            res.data = finalizerResult;
+          try {
+            const finalizerResult = await finalizerFn(
+              res.data,
+              functionParams,
+              params
+            );
+
+            // TODO: This is a bad check because the result could be falsy:
```
Danziger, line 135: Discuss this with @7i7o

src/api/background/handlers/storage/auth-state-change/auth-state-change.handler.ts
```diff
@@ -18,4 +19,5 @@ export async function handleAuthStateChange({
   const connected = await app.isConnected();
 
   await updateIcon(connected);
+  createContextMenus(connected);
```
Danziger, line 22: Everywhere else where there's an `updateIcon`, there's also a `createContextMenus`.

src/api/modules/connect/connect.background.ts
```diff
@@ -65,11 +65,12 @@ const background: BackgroundModuleFunction<void> = async (
     // add features available after connection
     await updateIcon(true);
     createContextMenus(true);
-  } catch (e: any) {
+  } catch (err) {
     await updateIcon(false);
     createContextMenus(false);
 
-    throw new Error(e);
+    // TODO: Replacing this with `throw err` produces better stack traces:
```
Danziger, line 72: @7i7o See here. Maybe something worth changing (in all methods).

src/api/modules/sign/chunks.ts
```diff
@@ -33,41 +35,64 @@ const chunks: {
...
       type: "chunk",
-      app: "wander",
-      data: chunk,
-      callID
+      data: chunk
     };
 
-    // send message
-    window.postMessage(message, window.location.origin);
+    if (import.meta.env?.VITE_IS_EMBEDDED_APP === "1") {
```
Danziger: This can be implemented better by hiding this conditional inside `isomorphicSendMessage`. I'll create a refactoring task for later.

src/api/modules/sign/utils.ts
```diff
@@ -47,7 +47,8 @@ export async function arconfettiIcon(): Promise<string | false> {
  */
 export async function calculateReward({ reward }: Transaction) {
   // fetch fee multiplier
-  const multiplier = await getSetting("fee_multiplier").getValue<number>();
+  const multiplier =
```
Danziger, line 50: This was sometimes coming back as `null`, causing `tx.reward` to be `NaN`.

src/components/popup/home/Balance.tsx
```diff
@@ -95,37 +125,43 @@ export default function Balance() {
...
+            transform: "translate(-50%, -50%)",
+            width: "20px",
+            height: "20px"
+          }}
+        />
+      ) : null}
+
+      <BalanceWrapper $hideBalance={hideBalance || isLoading}>
+        <BalanceText onClick={() => setHideBalance((val) => !val)} noMargin>
+          <NumberFlow
```
Danziger, line 143: By keeping the `NumberFlow` element always rendered, we get nicer transitions when the price is loaded, hidden or shown. While loading or hidden, it now renders like this, and selection is not possible:

image

This also matched the SDK button, which also blurs its content (which is also replaced by a fake value of `888.88` to avoid leaking the balance to the dApp:

image

src/components/popup/home/Balance.tsx
```diff
@@ -37,6 +36,45 @@ export default function Balance() {
     false
   );
 
+  useEffect(() => {
+    // TODO: The balance and fiat balance should be loaded and calculated from a provider / background service. Relying
```
Danziger, line 40: See here. Right now we have many components handling business logic, which causes issues like this where the functionality / data we need is inside a component, which means that it's only available when that component is rendered, and that in order to "subscribe" to that data we end up adding logic inside the component that doesn't belong there (separation of concerns, single responsibility). 

This is ok-ish for now, but once we have proper state management with Mobx, things like this should be moved there.

wander-embedded-sdk/src/components/button/wander-button.component.ts
```diff
@@ -280,11 +282,13 @@ export class WanderButton {
   }
 
   setBalance(balanceInfo: BalanceInfo) {
-    const formattedBalance = new Intl.NumberFormat(undefined, {
-      currency: balanceInfo.currency
-    }).format(balanceInfo.amount);
+    if (balanceInfo.amount === null) {
```
Danziger, line 285: If the balance is hidden in the app, it's also hidden here. Also, the real amount is not send, neither in `amount` nor in `formattedBalance`.

wander-embedded-sdk/src/components/button/wander-button.template.ts
```diff
@@ -64,6 +64,15 @@ export const getWanderButtonTemplateContent = ({
     transition: transform linear 50ms;
   }
 
+  .button::after {
```
Danziger, line 67: The clickable area now better match the button's visible shape:

image

wander-embedded-sdk/src/components/iframe/wander-iframe.component.ts
```diff
@@ -73,18 +73,19 @@ export class WanderIframe {
...
-      // If a single value is passed, we use it for default and auth-requests. Anything else fallbacks to the default
-      // (currently modal):
+      // If a single value is passed, we use it for default, settings and auth-request. auth and account fallbacks to
+      // the default (currently modal):
 
       const defaultLayoutConfig = WanderIframe.getLayoutConfig(routeLayout);
 
       this.routeLayout = {
         default: defaultLayoutConfig,
+        settings: defaultLayoutConfig,
```
Danziger, line 83: By default, Settings now also uses the popup layout, or the same one defined for "default" if only one custom layout option is passed (`else` block below).

wander-embedded-sdk/src/components/iframe/wander-iframe.component.ts
```diff
@@ -247,19 +248,20 @@ export class WanderIframe {
...
       }
 
       case "popup": {
         const position = layoutConfig.position || "bottom-right";
         this.wrapper.dataset.position = position;
-
-        cssVars.preferredWidth ??= layoutConfig.fixedWidth || routeConfig.width;
-        cssVars.preferredHeight ??=
-          layoutConfig.fixedHeight || routeConfig.height;
+        // Popup should not resize to fit content:
```
Danziger, line 262: Bug, I never intended it to work like it did, but I probably just copy-pasted the code from `modal`.

wander-embedded-sdk/src/components/iframe/wander-iframe.component.ts
```diff
@@ -247,19 +248,20 @@ export class WanderIframe {
 
     switch (layoutConfig.type) {
       case "modal": {
-        cssVars.preferredWidth ??= layoutConfig.fixedWidth || routeConfig.width;
```
Danziger, line 250: The `??` are incorrect, as that means that if a value for a CSS var is defined in `options.cssVars`, neither the layout config nor the route will be able to remove it, but that's desired in som cases, like the `popup` layout which might simply fallback to the default values.

src/wallets/router/router.utils.ts
```diff
@@ -37,18 +37,6 @@ export function isNavigateAction(
   return typeof to === "number" || !to.startsWith("/");
 }
 
-export function prefixRoutes(
```
Danziger, line 40: Not used.

wander-embedded-sdk/src/components/iframe/wander-iframe.template.ts
```diff
@@ -34,9 +34,9 @@ export const getWanderIframeTemplateContent = ({
       border-radius: var(--borderRadius, 10px);
       box-shadow: var(--boxShadow, 0 0 16px 0 rgba(0, 0, 0, 0.125));
       width: calc(var(--preferredWidth, 400px) - 2 * var(--borderWidth, 2px));
-      height: calc(var(--preferredHeight, 600px) - 2 * var(--borderWidth, 2px));
-      min-width: 400px;
-      min-height: 400px;
+      height: calc(var(--preferredHeight, 800px) - 2 * var(--borderWidth, 2px));
+      min-width: calc(400px - 2 * var(--borderWidth, 2px));
```
Danziger, line 38: Account for border width.

wander-embedded-sdk/src/wander-embedded.ts
```diff
@@ -309,12 +301,32 @@ export class WanderEmbedded {
         break;
 
       case "embedded_request":
-        const { pendingRequests } = message.data;
+        // Scenarios to test:
```
Danziger, line 302: Not working properly thought. I might create a new issue to fix it.

src/routes/popup/settings/index.tsx
```diff
@@ -208,7 +209,14 @@ export function MenuView({ params }: QuickSettingsViewProps) {
             >
               {browser.i18n.getMessage("cancel")}
             </Button>
-            <Button fullWidth onClick={removeDecryptionKey}>
+            <Button
```
Danziger, line 212: It's now possible to sign out from Settings.

src/utils/auth/auth.utils.ts
```diff
@@ -136,7 +139,14 @@ export async function createAuthPopup(
     const hasWallets = activeAddress && wallets.length > 0;
 
     if (!hasWallets) {
-      openOrSelectWelcomePage(true);
+      if (import.meta.env?.VITE_IS_EMBEDDED_APP === "1") {
+        postEmbeddedMessage({
```
Danziger, line 143: If the dApp tries to use `window.arweaveWallet` before the user is authenticated or a wallet is loaded and activated (this last part will not be required later, in order to make it load faster), the popup opens automatically.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1005,6 +1006,13 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
 
       setEmbeddedContextState(EMBEDDED_CONTEXT_INITIAL_STATE);
 
+      postEmbeddedMessage({
+        type: "embedded_auth",
+        data: {
+          userDetails: !userId || !session ? null : {}
```
Danziger, line 1012: @7i7o Any user details we might want to share with the dApp (as we did with Othent)?

src/utils/embedded/embedded.utils.ts
```diff
@@ -65,14 +69,31 @@ export function getEmbeddedAncestorOrigin() {
...
   try {
     const supabase = await getSupabaseClient();
     await supabase.auth.signOut();
-    window.location.href = "#/";
-    window.location.reload();
   } catch (err) {
     console.error("Error signing out:", err);
+
+    window.location.href = "#/";
+    window.location.reload();
```
Danziger, line 96: @pawanpaudel93 I've changed this a bit to avoid reloading the page unless there's an error. Instead, the storage is cleared. This is because I'm also using this function when the user manually signs out, and I want the transition between being authenticated and not being authenticated to be nicer.

src/utils/icon.ts
```diff
@@ -95,6 +96,13 @@ const lockedLogos: LogosByEnvironment = {
 export async function updateIcon(hasPerms: boolean) {
   const val = await ExtensionStorage.get("decryption_key");
 
+  if (import.meta.env?.VITE_IS_EMBEDDED_APP === "1") {
+    postEmbeddedMessage({
```
Danziger, line 100: Embed has no icon, but it has the connection indicator, which has similar states. There's no "locked" state though.

src/utils/messaging/strategies/iframe/iframe-messaging.strategy.ts
```diff
@@ -0,0 +1,311 @@
...
+let messageCounter = 0;
+
+function getPostMessageFunction<K extends MessageID>(
+  messageData: MessageData<K>
+): () => Promise<ReturnType<OnMessageCallback<K>>> {
+  let postMessageTargetOrigin = "";
+
+  const { destination, messageId, data } = messageData;
+
+  // TODO: isInsideIframe is an incorrect check because we might be running the app standalone.
```
Danziger, line 38: The app won't run standalone once this is merged until this is addressed.

src/api/modules/sign/utils.ts
pawanpaudel93: @Danziger I think I have fixed issue with `getSetting` not returning default value if nothing is stored in my current open PR. Plasmo storage returns `undefined` if it doesn't find a key and StorageMock was returning `null` (as it uses sessionStorage internally) so the `getSetting` was not returning the default value as it checks if the returned value is `undefined` to return the default value.

pawanpaudel93:
@Danziger While testing the sign and dispatch api in playground, the tags for the transactions seems to be missing from the popup.

Issue: https://communitylabs.atlassian.net/browse/ARC-1208

pawanpaudel93:
We should probably store the connected apps to localStorage and I can update that in my PR where I have created two instances of storage (one for session and another for local)

pawanpaudel93:
The button displaying the connection status (green or gray dot) works correctly when using the API to connect or disconnect. However, if an app is removed from "Connected Apps" in settings or disconnected from the wallet dashboard, the button does not update to reflect the disconnected status.

Issue: https://communitylabs.atlassian.net/browse/ARC-1209

image

pawanpaudel93:
When connecting to an app if logo is not passed in the connect API then no logo is shown. In BE favicon is used if no logo is passed. We could do something like this for Embed if no logo is passed to the connect API.
```js
function getAppLogo(): string | null {
  const getAbsoluteUrl = (url: string): string | null => {
    if (!url) {
      return null;
    }
    try {
      return new URL(url, window.location.origin).href;
    } catch {
      return null;
    }
  };

  const faviconSelectors = [
    "link[rel='icon'][type='image/svg+xml']",
    "link[rel='icon'][sizes='32x32']",
    "link[rel='icon'][sizes='48x48']",
    "link[rel='icon'][sizes='96x96']",
    "link[rel='icon']",
    "link[rel='shortcut icon']",
    "link[rel='apple-touch-icon'][sizes='180x180']",
    "link[rel='apple-touch-icon']",
    "link[rel='apple-touch-icon-precomposed']",
    "meta[property='og:image']",
    "img[alt*='logo']",
    "img[src*='logo']"
  ];

  for (const selector of faviconSelectors) {
    const element = document.querySelector(selector);
    const url =
      element instanceof HTMLLinkElement ? getAbsoluteUrl(element.href) : null;
    if (url) {
      return url;
    }
  }

  const defaultFavicon = `${window.location.origin}/favicon.ico`;
  return defaultFavicon;
}
```

7i7o:
> In BE favicon is used if no logo is passed. We could do something like this for Embed if no logo is passed to the connect API.

Let's add this to a new ticket so we can move forward and merge the pending PRs for embed 👍

7i7o (approved):
tACK 👍

PR #558, opened 2024-12-03
fix: handle new contact navigation from transaction details

pawanpaudel93:
## Summary

This PR resolves the issue with navigation from transaction details to new contact screen.


7i7o (approved):
tACK 👍

PR #572, opened 2024-12-06
Bug: Fix gateway issues

Danziger:
Fix gateway-related issues that log messages to the background script's Console.

### DIFF
--- a/src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
+++ b/src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
@@ -3,8 +3,9 @@ import { type Alarms } from "webextension-polyfill";
 import { AOProcess } from "~lib/ao";
 import { AO_ARNS_PROCESS } from "~lib/arns";
 import type {
-  GatewayAddressRegistryCache,
-  ProcessedData
+  GatewayAddressRegistryItem,
+  GatewayAddressRegistryItemData,
+  PaginatedResult
 } from "~gateways/types";
 import {
   RETRY_ALARM,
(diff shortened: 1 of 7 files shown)


src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
```diff
@@ -22,22 +23,26 @@ export async function handleGatewayUpdateAlarm(alarm?: Alarms.Alarm) {
...
-  let procData: ProcessedData[] = [];
+  let garItemsWithChecks: GatewayAddressRegistryItem[] = [];
 
   try {
     // fetch cache
     const ArIO = new AOProcess({ processId: AO_ARNS_PROCESS });
-    const gateways = (await ArIO.read({
+
+    const gatewaysResult = await ArIO.read<
+      PaginatedResult<GatewayAddressRegistryItemData>
```
Danziger, line 33: This result is paginated so the actual "gateways array" is nested and the previous enforced type was incorrect. Not sure why that might be, because I guess someone would have noticed if this never worked?

src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
```diff
@@ -22,22 +23,26 @@ export async function handleGatewayUpdateAlarm(alarm?: Alarms.Alarm) {
...
-    const garItems = extractGarItems({ gateways });
+    });
+
+    const garItems = extractGarItems(gatewaysResult.items);
 
     // healtcheck
-    await pingUpdater(garItems, (newData) => {
-      procData = [...newData];
+    await pingUpdater(garItems, (nextGarItemsWithChecks) => {
+      garItemsWithChecks = nextGarItemsWithChecks;
```
Danziger, line 42: Don't like this, but whatever... We can refactor it later.

src/gateways/types.ts
```diff
@@ -1,67 +1,92 @@
...
+  status: "unknown";
+}
+
+export interface GatewayPendingCheck {
+  status: "pending";
 }
 
-export interface GatewayAddressRegistryCache {
-  // contractTxId: string;
-  gateways: Record<string, GatewayAddressRegistryItemData>;
```
Danziger, line 28: Types were mostly incorrect, incomplete or inaccurate. For instance, this structure never actually existed.

src/gateways/wayfinder.ts
```diff
@@ -8,24 +8,31 @@ import { EventType, trackEvent } from "~utils/analytics";
...
-  if (!wayfinderEnabled && requirements.arns) {
-    return {
-      host: "arweave.dev",
-      port: 443,
-      protocol: "https"
-    };
-  }
+  // This should have been loaded into the cache by handleGatewayUpdateAlarm, but sometimes this function might run
+  // before that, so in that case we fall back to the same behavior has having the Wayfinder disabled:
+  const procData = await getGatewayCache();
```
Danziger, line 16: Before, `procData` was `undefined` because the alarm handler that had to update the gateway cache would throw an error. I've added a check in case that happens again for whatever reason, but the original issues are fixed.

7i7o (approved):
utACK 👍

PR #574, opened 2024-12-09
feat: adds default app icon placeholder

### LINKED ISSUE #570: Auth: Add a default app icon (e.g. lock icon) to `HeadV2`
In `src/components/popup/HeadV2.tsx` see:

```
  const appName = appInfo?.name || url;
  const appIconPlaceholderText = appName?.slice(0, 2).toUpperCase();
```

Note that if `url` is used as a fallback, we should:
- Get the first 2 letters from the domain, not from the subdomain, on a regular site.
- Get the first 2 letters from the subdomain if using a gateway.

It might just be simpler to use a defalt icon (e.g. lock icon) if there's no `appName` available.

Jonathansoufer:
## Description

This PR aims to implement a reusable way to handle Gateway's URLs and extract placeholder characters to those APPs lacking logos/favicons.

## Related issues

Fixes: [570](https://github.com/arconnectio/ArConnect/issues/570)

## Manual testing steps
N/A

### After


src/utils/urls/getAppIconPlaceholder.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (appName) {
+    return appName.slice(0, 2).toUpperCase();
+  }
+
+  if (appInfo.url) {
+    try {
+      const parsedUrl = new URL(appInfo.url);
+      const hostnameParts = parsedUrl.hostname.split(".");
+
+      if (hostnameParts.length > 2) {
```
Danziger: This will not work reliably:

- ✅ If `hostname` is `github.com`, it returns `GI`.
- ❌ If `hostname` is `subdomain.github.com`, it returns `SU`.
- ✅ If `hostname` is `bazar.arweave.dev`, it returns `BA`.

The most reliable way to do this is to always take the base domain (e.g. `github.com`, `arweave.dev`...) and make a `GET` request. Gateways return a JSON like this (from `https://[arweave.dev](https://arweave.dev/)/`):

```
{
    "version": 5,
    "release": 74,
    "queue_length": 0,
    "peers": 512,
    "node_state_latency": 0,
    "network": "arweave.N.1",
    "height": 1565184,
    "current": "fmOLnss-H7UnqjyexJctWE3BYr_aHTKHIhFS4MBHDnaPHPxCfLKQDh7z4OdGgCHh",
    "blocks": 1565185
}
```

I would start with adding a new interface for this new data you need:

```
export interface AppLogoInfo extends AppInfo {
  type: "default" | "gateway"; // We could add more like "githubPages"
  placeholder: string;
}
```

Then, in `HeadAuth` change:

```
const [appInfo, setAppInfo] = useState<AppInfo>(appInfoProp);
```

to:

```
const [appLogoInfo, setAppLogoInfo] = useState<AppLogoInfo>(appInfoProp);
```

And update the `useEffect` below to instantiate `Application`, asynchronously check if the application URL is a gateway or not and update the state accordingly.

You then also need to change `HeadV2`'s to work with the new data.

You should probably also update `SquircleImg` to have a loading state that will be displayed while the gateway checks are still running. Simply making the background pulsate would be ok, or otherwise there's a spinner in `import { Loading } from "@arconnect/components"`.

There's no need for the "lock" fallback in this case.

src/utils/urls/getAppIconPlaceholder.ts
```diff
@@ -0,0 +1,40 @@
...
+  url: string
+): Promise<AppLogoInfo | undefined> {
+  try {
+    const { hostname } = new URL(url);
+
+    const parts = hostname.split(".");
+
+    const baseDomain = parts.length > 2 ? parts.slice(-2).join(".") : hostname;
+
+    const isGatewayUrl = await isGateway(url);
```
Danziger: `url` here is the URL of the connected dApp, so this will not be sending the GET request to the right place. For example, if `url = "https://bazar.arweave.dev/"`, the gateway URL is `https://arweave.dev`.

I would change these first few lines to:

```
const parts = hostname.split(".");
const baseDomain = parts.slice(-2).join(".");
const candidateGatewayURL = parts.length > 1 ? parts.slice(1).join(".") : null;
const isGatewayURL = !!candidateGatewayURL  && await isGateway(candidateGatewayURL);
```

Jonathansoufer: Thank you. Good catch. Addressed.

Danziger (approved):

PR #577, opened 2024-12-12
fix: add transaction completion page and use multiple gateways for tx query

pawanpaudel93:
## Summary

This PR improves transaction loading by adding multiple gateway support with retries and introduces a transaction completion page aligned with the mobile app.


7i7o (approved):
tACK 👍

PR #578, opened 2024-12-12
refactor: update sign header, humanize timestamps, hide fiat amount

pawanpaudel93:
## Summary

This PR adds improvements to the sign transaction popup:
- Update sign header for AO transactions with `Action` tag if present.
- Humanize some known timestamp tags
- Hide fiat amount if its zero.


7i7o (approved):
tACK 👍

PR #579, opened 2024-12-13
refactor: update the default assets list

pawanpaudel93:
## Summary

This PR updates the default assets list and makes AO native token import faster with cached Info.


7i7o (approved):
utACK 👍

nicholaswma (approved):

PR #582, opened 2024-12-16
Feature: `decrypt()` & `signMessage()` permissions + fix `dispatch()` requesting permissions twice (when rejected)

Danziger:
- `decrypt()` now always asks permissions.
- `signMessage()` now always asks permissions.
- `dispatch()` won't re-request permissions / re-attempt posting if the user rejected the auth request (it will still do it if the auth request is accepted but uploading it to Turbo fails).

Partially fixes https://github.com/arconnectio/ArConnect/issues/550.
Closes https://github.com/arconnectio/ArConnect/issues/551.

### DIFF
--- a/src/api/modules/decrypt/decrypt.background.ts
+++ b/src/api/modules/decrypt/decrypt.background.ts
@@ -20,8 +20,6 @@ const background: BackgroundModuleFunction<string | Uint8Array> = async (
   // validate data
   isRawArrayBuffer(data);
 
-  const message = Object.values(data);
-
   // override with byte array
   data = new Uint8Array(Object.values(data));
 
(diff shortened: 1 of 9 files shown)


src/api/modules/decrypt/decrypt.background.ts
```diff
@@ -110,18 +114,30 @@ const background: BackgroundModuleFunction<string | Uint8Array> = async (
...
+    decryptedData = new Uint8Array(decrypted);
   } else {
     // remove wallet from memory
     freeDecryptedWallet(privateKey);
 
     throw new Error("Invalid options passed", options);
   }
+
+  // request "decrypt" popup
+  await requestUserAuthorization(
```
Danziger, line 132: With this change, we would be ALWAYS asking permissions to decrypt data.

This might be an issue but until we rework allowances, I don't think there's a better way. One option could be to do the following check, just like we do with `sign()`:

```
const alwaysAsk = allowance.enabled && allowance.limit.eq(BigNumber("0"));
```

But the issue I have with this is that setting an allowance limit doesn't mention anything to the user about decrypting data (or signing messages).

7i7o: I agree. The copy is not the best for this use cases
Let's go with this approach first, and re-asses later on when allowances has its update

7i7o (approved):
tACK 👍

PR #586, opened 2024-12-18
Adds an Error Boundary capability

### LINKED ISSUE #559: Routing: Review `return <></>` instances
Some components have checks where they do `return <></>`.

Chances are the right thing to do would be to either show a loader/spinner or do `return <Redirect to="..." />` instead.

Most of these instances are marked with a `TODO` comment: `// TODO: Should this be a redirect?`

Jonathansoufer:
## Description

This PR aims to implement an Error Boundary capability to handle exceptions/errors through the application.

## Related issues

Fixes: [559](https://github.com/arconnectio/ArConnect/issues/559)

## Screenshot

Production Environment
Screenshot 2024-12-18 at 16 56 43

Development Environment
Screenshot 2024-12-18 at 16 56 32

## Manual testing steps
N/A

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,0 @@
-PLASMO_PUBLIC_APP_TYPE=extension
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -1,4 +1,8 @@
 {
+  "fallback": {
+    "message": "Oops, something isn't right. Try again later.",
+    "description": "Fallback error message"
+  },
   "extensionDescription": {
     "message": "Secure wallet management for Arweave",
     "description": "Extension description"
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -1349,13 +1349,31 @@
     "message": "Sign items",
     "description": "Batch sign message popup title"
   },
+  "titles_decrypt": {
+    "message": "Decrypt",
+    "description": "Request decrypt authorization password"
+  },
+  "decrypt_description": {
+    "message": "$APPNAME$ wants to decrypt and access the following data:",
+    "description": "Description for decrypting some data",
+    "placeholders": {
+      "appname": {
+        "content": "$1",
+        "example": "permafacts.arweave.dev"
+      }
+    }
+  },
+  "decrypt_authorize": {
+    "message": "Authorize",
+    "description": "Authorize button label"
+  },
   "titles_signature": {
     "message": "Sign message",
     "description": "Sign message popup title"
   },
   "sign_data_description": {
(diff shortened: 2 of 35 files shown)


src/utils/error/error.utils.ts
```diff
@@ -1,3 +1,17 @@
...
 }
+
+export enum ErrorTypes {
+  Error = "Error",
+  RangeError = "RangeError",
+  ReferenceError = "ReferenceError",
+  SyntaxError = "SyntaxError",
+  URIError = "URIError",
+  PageNotFound = "Page not found",
+  SettingsTypeNotFound = "Settings type not found",
```
Danziger: I would avoid labelling all of them as "not found", as there are some subtle differences in these ones:

- `SettingsTypeNotFound`: I would do something like `throw new Error(settings.type ? ErrorTypes.MissingSettingsType : ErrorTypes.UnexpectedSettingsType)`. It would even be better if you throw a custom error where we can add additional details (the actual value we got, the expected value(s)...) but that's probably overkill right now.
- `TxIdNotFound`: I would rename this one to `MissingTxId`.

Like this they convey a bit more information:
- "Not found": Something was queried/mapped/loaded but was not there.
- "Missing": Something should have been provided/set, but wasn't.
- "Unexpected": Something was provided, but the value wasn't right.

.env.example
```diff
@@ -1 +0,0 @@
-PLASMO_PUBLIC_APP_TYPE=extension
```
Danziger, line 1: This one wasn't added by mistake, it's just here so that when new devs copy the `.env` file, they get the default build behavior (building the browser extension).

assets/_locales/en/messages.json
```diff
@@ -1,4 +1,8 @@
 {
+  "fallback": {
```
Danziger, line 2: Can you please also add this to the other languages. Just the same value for now (English), someone will go throw all of them before each release.

src/components/page/common/Fallback/fallback.view.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from "styled-components";
+import browser from "webextension-polyfill";
+
+import { Text } from "@arconnect/components";
+
+export const FallbackView = () => {
+  return (
+    <DivWrapper>
+      <Text heading>{browser.i18n.getMessage("fallback")}</Text>
+    </DivWrapper>
```
Danziger, line 39: If `process.env.NODE_ENV === "development"` only, could you add a second line with details about the error: name, message, stack.

And in all envs, it would be good to have a button to reload the app (and redirect to the root).

Danziger (approved):

Danziger:
uACK 👍

PR #600, opened 2025-01-02
fix: return default gateway if no gateways found with wayfinder disabled

pawanpaudel93:
## Summary

This PR fixes the issue with `findGateway` to return a default gateway in case of error or no gateway found. 

## How to Test
1. Disable the wayfinder and try to send ao tokens in `development` or prod Arconnect. You will see `Could not send transaction error` error.
2. Checkout `arc-884/wayfinder-disabled-fix` and try to send ao tokens again and make sure you can send tokens successfully without `Could not send transaction error`.


7i7o (approved):
tACK 👍

PR #603, opened 2025-01-03
fix: reset import issue

pawanpaudel93:
## Summary

This PR resolves the issue of loading the default tokens after reset and onboarding. Also, it updates `signDataItem` popup to show action as header and hide fiat amount if absent.

## How to Test
1. Reset the wallet
2. Generate or import wallet using keyfile/seedphrase
3. Make sure there are default tokens present


7i7o (approved):
tACK 👍

PR #623, opened 2025-01-23
feat: Embedded SDK

Danziger:
**Setup:**

- Clone this repo.
- `yarn install`
- `yarn build:wallet-api && yarn dev:iframe` - The SDK imports some shared dependencies (`setupWalletSDK()`) from an already built version of the main repo.
- `cd wander-embedded-sdk/`
- `yarn dev` - This is the build of the SDK itself.
- Clone https://github.com/[kranthicodes/WE-SDK](https://github.com/kranthicodes/WE-SDK).
- Update `App.tsx` with:

```
  const [message, setMessage] = useState("");
  const [instance, setInstance] = useState<WanderEmbedded | null>(null);

  useEffect(() => {
    const wanderInstance = new WanderEmbedded({
      iframe: {
        routeLayout: {
          auth: "popup",
        },
      },
    });
    
    setInstance(wanderInstance);
  }, []);

  const handleSignMessage = async () => {
    await (window.arweaveWallet as any)?.connect(["SIGNATURE"]);
    await (window.arweaveWallet as any)?.signMessage(new TextEncoder().encode(message));
  };
```

- Replace `wander-embedded-sdk` in `package.json` with: `"wander-embedded-sdk": "link:./../wander/Wander/wander-embedded-sdk/"`
- `pnpm install`
- `pnpm dev`

**TODO:**

The tasks that are prefixed with `TODO: ` can be found as an actual `TODO` in the code.

  -  Update `WanderIframe` component to use shadow DOM and set its `config` property (derived from `options`, including default values), just like in `WanderButton`. This also means there's no need to use default values in CSS `var()` anymore (again, just like in `WanderButton`).

  - Include media queries for viewport size in `WanderIframe`. On mobile, take the whole screen or use a drawer. Both options could be fully expanded or have padding around. We probably need a different set of options for mobile. Make sure the button is correctly positioned in all layouts / screen sizes.

image

  - In `interface WanderEmbeddedModalCSSVars` there are some properties defined under a comment "App wrapper (inside iframe):". Those should be passed to the iframe context and used from there.
  - We need a wrapper around the `iframe` to be able to implement an animation like [Passkeys](https://passkeys.foundation/playground), as otherwise, if we resize the iframe directly, its content is affected.
  - Add close button inside iframe and make sure the spinner shows straight away.
  - There should be a method to change options/config (use case is changing the theme or button labels (language) based on user preference) after instantiation.
  - TODO: Animate/transition this. First close the old layout. Then open the new one.
  - Add option to configure the size-images based on route on the side-by-side view (or send them from the modal)
  - "popup" layout should probably not resize, only modal.
  - TODO: Pass theme, balance config and max width/height to iframe:
  - Add customizable default size for each layout.
  - Handle all / most other `TODOs` in this PR's code.
  - Make sure the SDK class cannot be instantiated twice (e.g. turn it into a singleton, throw an error or destroy the previous instance before creating a new one).

**TODO (maybe later):**

  - Add different default themes: sketch, smooth, rounded (current one).
  - Considering having different transitions or a CSS variable for transition duration / speed.
  - Add effect when spending/signing


shim.d.ts
```diff
@@ -60,13 +60,10 @@ declare module "@arconnect/webext-bridge" {
...
-    embedded_open: any;
-    embedded_close: any;
-    embedded_resize: any;
-    // TODO: Maybe yes, maybe not (we should not drift too far away from the BE wallet API)
-    // embedded_auth: any;
-    // embedded_balance: any;
-    // embedded_info: any;
+    embedded_auth: EmbeddedAuthMessageData;
+    embedded_balance: EmbeddedBalanceMessageData;
+    embedded_resize: EmbeddedResizeMessageData;
```
Danziger, line 65: I've removed the "open" message as those would always be followed by a "resize" one, so having "resize" already signals to the SDK that the app should open.

src/components/page/page.component.tsx
```diff
@@ -1,24 +1,82 @@
...
+            height
+          }
+        });
+
+        // For debugging only:;
+        setHeight(height);
+      }
+    });
+
+    resizeObserver.observe(mainElement);
```
Danziger, line 59: This will be notifying the parent about any change in size due to images loading, elements appearing/disappearing, etc.

src/utils/embedded/utils/messages/embedded-messages.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+  data: EmbeddedMessageMap[K];
+}
+
+export function postEmbeddedMessage<K extends EmbeddedMessageId>({
+  type,
+  data
+}: PostEmbeddedMessageData<K>) {
+  if (!type.startsWith("embedded_"))
+    throw new Error(
+      `Only "embedded_auth", "embedded_balance", "embedded_resize" or "embedded_data" can be posted to the upper frame.`
```
kranthicodes: `embedded_data` is this intentional ? embedded_close is missing i guess based on map in types file

Danziger: My bad, the last should be `embedded_close`. I've pushed commit with the correct code.

shim.d.ts
7i7o: 👍 
We need to explain this thoroughly for devs in the docs

Do you think we might need a "embedded_focus" or should we just use resize for something like that?

7i7o (approved):
ACK 👍

shim.d.ts
Danziger: I think we can just add an `isFocused` property to the `embedded_resize` event data. It might be more convenient to have all the information in a single event when you need to open/resize the iframe modal/popup rather than getting that information from 2 separate events.

PR #631, opened 2025-01-27
feat: Embedded SDK 

kranthicodes:
## Summary

- separate vite config to build wallet-api into sdk directory
  `build:wallet-api": "vite build --config vite.wallet-config.js`
- build setup for Embedded SDK using tsup
- options to customize button and iframe styles
- wallet api communicates with injected iframe via postMessage
- proxy wallet api
- Lite readme docs


README.md
```diff
@@ -0,0 +1,135 @@
...
+| `iframeStyles` | `object`                     | `undefined` | Inline styles for the iframe (e.g., `{ border: 'none', borderRadius: '12px' }`).                                    |
+| `logo`         | `string`                     | `undefined` | URL for a custom logo displayed on the Wander button.                                                               |
+| `balance`      | `string`                     | `undefined` | Show balance information on the button.                                                                             |
+| `onOpen`       | `() => void`                 | `undefined` | Callback fired when the Wander iframe is opened.                                                                    |
+| `onClose`      | `() => void`                 | `undefined` | Callback fired when the Wander iframe is closed.                                                                    |
+| `onResize`     | `(data: ResizeData) => void` | `undefined` | Callback fired when the Wander iframe is resized. The `data` object typically includes new width and height values. |
```
Danziger, line 46: `ResizeData` should be:

```
export interface EmbeddedResizeMessageData {
  routeType: RouteType;
  preferredLayout: EmbeddedLayout;
  width?: number;
  height: number;
}
```

`onAuth(userDetails: UserDetails | null)` is missing.

I also forgot to add `onNotification(notificationCount: number)`. The corresponding message ID would be `embedded_notification: EmbeddedNotificationMessageData { notificationCount: number }`, and it should display a small notification counter/alert on/next to the button.

README.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Methods
+
+#### `open()`
```
Danziger, line 50: We are probably missing the props: `isOpen`, `isAuthenticated`, `width`, `height`.

README.md
```diff
@@ -0,0 +1,135 @@
...
+## API Reference
+
+### Constructor Options
+
+The `WanderEmbedded` constructor accepts an optional configuration object with the following properties:
+
+| **Option**     | **Type**                     | **Default** | **Description**                                                                                                     |
+| -------------- | ---------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------- |
+| `buttonStyles` | object \| 'none'             | `undefined` | Customize the button’s appearance: - **object**: inline style object ; **'none'**: hide the button                  |
+| `iframeRef`    | `HTMLIFrameElement`          | `undefined` | Use an existing iframe element instead of creating a new one.                                                       |
```
Danziger, line 40: I would combine/nest all props to avoid the illusion of having so many options, which in reality cannot be used together:

```
 iframe?: IFrameOptions | HTMLIFrameElement ;
 button?: ButtonOptions | false; 

[...]

interface IFrameOptions {
  id?: string;
  className?: string | Record<StateModifier , string>;
  cssVars?: IFrameCSSVars | Record<StateModifier, IFrameCSSVars>;
  routeLayout: Record<RouteType, EmbeddedLayout>;
}

interface ButtonOptions {
  id?: string;
  className?: string | Record<StateModifier , string>;
  cssVars?: ButtonCSSVars | Record<StateModifier, ButtonCSSVars>;
  logo?: boolean | string;
  balance?: boolean | BalanceOptions;
  position: ButtonPosition;
}

interface BalanceOptions {
  balanceOf: "total" | string; // string would be a token id
  currency: "auto" | string; // "auto" would be the one the user selected on the wallet, string would be a token id or currency symbol (e.g. USD).
}

type StateModifier = "isAuthenticated" | "isOpen" | "isAuthRoute" | "usAccountRoute" | "isSettingsRoute" | "isAuthRequestRoute" | "isDefaultRoute";

inteface IFrameCSSVars {
  // TODO: Similar to just using stlyes but styes might cause issues with the different states/transitions of the iframe. Also, we probably don't want to give such fine-grained control to developers. In that case, just provide the iframe yourself.
}

interface ButtonCSSVars {
  // TODO: Similar to just using stlyes but styes might cause issues with the different states/transitions of the button. Also, we probably don't want to give such fine-grained control to developers. In that case, just create and manage the button yourself.
}

type ButtonPosition = "top-right" | "bottom-right" | "top-left" | "bottom-left";
   
```

Even with these changes, sites that use allow users to change the theme or to use the default (system) one will have issues implementing that functionality with this API, so some additional thought needs to go into this.

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -12,102 +12,134 @@ export function setupWalletSDK(targetWindow: Window = window) {
...
+        return resolve(res.data);
+      }
+    });
   }
 
+  /**
+   * Create the Proxy object.
+   */
+  const proxyWallet = new Proxy<Record<string, any>>(
+    {},
```
Danziger: We probably want to prepopulate this:

```
for (const mod of foregroundModules) {
  walletAPI[mod.functionName] = (...params: any[]) =>{
    return callForegroundThenBackground(mod.functionName, ...params);
  }
}
```

Then:

```
const proxyWallet = new Proxy<Record<string, any>>(walletAPI, { ... });
```

This way also the `Reflect.get(target, propKey)` will actually be used, and you don't need this inside the proxy:

```
// Pass through known constants or objects:
if (propKey === "walletName") return "ArConnect";
if (propKey === "walletVersion") return version;
if (propKey === "events") return events;
```

`walletName` should be `ArConnect Embedded` if `import.meta.env?.VITE_IS_EMBEDDED_APP === "1"`.

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -12,102 +12,134 @@ export function setupWalletSDK(targetWindow: Window = window) {
...
+      // execute foreground module
+      // TODO: Use a default function for those that do not have/need one and see if chunking can be done automatically or if it is needed at all:
+      const foregroundResult = mod ? await mod.function(...params) : null;
+
+      // 2. Prepare the payload to send
+      const callID = nanoid();
+      const data: ApiCall & { ext: "arconnect" } = {
+        type: `api_${mod.functionName}`,
+        ext: "arconnect",
+        callID,
```
Danziger, line 51: Adding the SDK version property here could be useful for Embedded.

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -12,102 +12,134 @@ export function setupWalletSDK(targetWindow: Window = window) {
...
+    return new Promise<any>(async (resolve, reject) => {
+      // execute foreground module
+      // TODO: Use a default function for those that do not have/need one and see if chunking can be done automatically or if it is needed at all:
+      const foregroundResult = mod ? await mod.function(...params) : null;
+
+      // 2. Prepare the payload to send
+      const callID = nanoid();
+      const data: ApiCall & { ext: "arconnect" } = {
+        type: `api_${mod.functionName}`,
+        ext: "arconnect",
```
Danziger: This value would need to change to `"arconnectEmbedded"` based on `import.meta.env?.VITE_IS_EMBEDDED_APP`

src/iframe/iframe.tsx
```diff
@@ -15,6 +15,31 @@ import { EmbeddedProvider } from "~utils/embedded/embedded.provider";
 export function ArConnectEmbeddedApp() {
   useEffect(() => {
     handleSyncLabelsAlarm();
+
+    window.addEventListener("message", (event: MessageEvent) => {
```
Danziger: This should probably be relocated to `setupBackgroundService` (but run only for Embedded based on ENV variable).

wander-embedded-sdk/src/WanderEmbedded.ts
```diff
@@ -0,0 +1,97 @@
...
+  private readonly DEFAULT_IFRAME_SRC = "http://localhost:5174/";
+  private options: WanderEmbeddedOptions;
+
+  constructor(options?: WanderEmbeddedOptions) {
+    this.options = options || {};
+    this.components = this.initializeComponents(options);
+  }
+
+  private initializeComponents(options?: WanderEmbeddedOptions): UIComponents {
+    const iframe = new WanderIframe({
```
Danziger: We should avoid creating this if `options.iframeRef` is provided:

```
if (!options?.iframeRef) {
  const iframe = new WanderIframe({ ... });
  document.body.appendChild(iframe.getElement());
}
```

Messages should be handled in this root class and decoupled from the component classes, which should only care about styling but should be controlled from outside.

wander-embedded-sdk/src/WanderEmbedded.ts
```diff
@@ -0,0 +1,97 @@
...
+      container.appendChild(button.getElement());
+      document.body.appendChild(container);
+
+      return { container, button, iframe };
+    }
+
+    return { iframe };
+  }
+
+  private createContainer(): HTMLDivElement {
```
Danziger: Why do we need the container?

wander-embedded-sdk/src/WanderEmbedded.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.components.iframe.hide();
+        this.options.onClose?.();
+        break;
+      case "embedded_resize":
+        this.components.iframe.resize(
+          message.data as IncomingResizeMessageData
+        );
+        this.components.iframe.show();
+
+        this.options.onOpen?.();
```
Danziger: We need to add a `isOpen` property and only call `onOpen` when `isOpen` goes from `false` to `true`.

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -12,102 +12,134 @@ export function setupWalletSDK(targetWindow: Window = window) {
...
-  };
-
-  /** Inject each module */
+  /**
+   * Store modules in a map for quick lookup.
+   */
+  const moduleMap = new Map<string, (typeof foregroundModules)[number]>();
   for (const mod of foregroundModules) {
-    /** Handle foreground module and forward the result to the background */
-    WalletAPI[mod.functionName] = (...params: any[]) =>
```
Danziger, line 25: Did you check if theses changes affect the BE API?

kranthicodes: I was able to spin up local BE setup and was able to connect to dapps without issues

wander-embedded-sdk/src/WanderEmbedded.ts
kranthicodes: you mean something like below?

```ts
case "embedded_resize":
        this.components.iframe.resize(
          message.data as IncomingResizeMessageData
        );
        this.components.iframe.show();

        if (!this.isOpen) {
          this.isOpen = true;
          this.options.onOpen?.();
        }
        this.options.onResize?.(message.data as IncomingResizeMessageData);
        break;
```

Danziger: I would also avoid calling `this.components.iframe.show()` if it's already being shown:

```
case "embedded_resize":
        this.components.iframe.resize(
          message.data as IncomingResizeMessageData
        );

        if (!this.isOpen) {
          this.isOpen = true;
          this.components.iframe.show();
          this.options.onOpen?.();
        }

        this.options.onResize?.(message.data as IncomingResizeMessageData);
        break;
```

PR #632, opened 2025-01-30
Add support for ArNS primary names

kunstmusik:
This PR introduces nameservice.ts that searches and provides ArNS Primary Name profiles in addition to ANS profiles. Code that formerly used ans.ts directly was refactored to use nameservice. Additionally, an in-memory cache backed by ExtensionStorage was introduced to optimize profile lookup.

PR #641, opened 2025-02-06
Add support for ArNS primary names

7i7o:
This PR introduces nameservice.ts that searches and provides ArNS Primary Name profiles in addition to ANS profiles. Code that formerly used ans.ts directly was refactored to use nameservice. Additionally, an in-memory cache backed by ExtensionStorage was introduced to optimize profile lookup.
Rebased #632 opened by @kunstmusic onto `development` to include the latest updates


7i7o:
Replaced by #665

PR #656, opened 2025-02-11
fix: removed token mirror

nicholaswma:
Additionally updated token_balance and user_tokens endpoints to not rely on the mirror

**Future Considerations:**
- Remove all references to `AO_NATIVE_TOKEN_BALANCE_MIRROR` `getNativeTokenBalance`


7i7o (approved):
tACK 👍

PR #661, opened 2025-02-13
fix: add getFormattedColor utility for color formatting

pawanpaudel93:
## Summary

This PR introduces a utility function, `getFormattedColor`, which ensures proper formatting of color values, including validating hex codes that start with `#`.  

Following Wander's rebranding, `primaryText` was changed from an RGB value to a hex code. However, `ThemeBackgroundObserver` did not recognize it as a valid color, preventing `ARCONNECT_THEME_TEXT_COLOR` from being set. As a result, components relying on `currentColor` such as `Loading` (which became invisible in light mode, affecting elements like dashboard balance loading, activity tab loading, etc.) were impacted.  

By incorporating `getFormattedColor`, we ensure consistent color detection, preventing such issues.


7i7o (approved):
tACK 👍

PR #665, opened 2025-02-14
Add support for ArNS primary names

7i7o:
This PR introduces nameservice.ts that searches and provides ArNS Primary Name profiles in addition to ANS profiles. Code that formerly used ans.ts directly was refactored to use nameservice. Additionally, an in-memory cache backed by ExtensionStorage was introduced to optimize profile lookup.
Rebased #632 opened by @kunstmusic onto development to include the latest updates

### DIFF
--- a/src/routes/auth/connect.tsx
+++ b/src/routes/auth/connect.tsx
@@ -39,6 +39,9 @@ import WanderIcon from "url:assets/icon.svg";
 import Image from "~components/common/Image";
 import { Flex } from "~components/common/Flex";
 import { svgie } from "~utils/svgies";
+import { useNameServiceProfile } from "~lib/nameservice";
+import { FULL_HISTORY, useGateway } from "~gateways/wayfinder";
+import { concatGatewayURL } from "~gateways/utils";
 
 type Page = "unlock" | "connect" | "permissions" | "confirm";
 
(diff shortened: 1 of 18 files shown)


nicholaswma (approved):
tACK - 
- this isn't directly related to this PR, but I'd feel like users should be able to adjust their Wander Account names, without restrictions if they have a profile. Maybe add a badge or indicator if they are using a profile name. (Would this be a problem with an address having multiple profiles too?)

7i7o:
> tACK -
> 
> * this isn't directly related to this PR, but I'd feel like users should be able to adjust their Wander Account names, without restrictions if they have a profile. Maybe add a badge or indicator if they are using a profile name. (Would this be a problem with an address having multiple profiles too?)

I agree on users having the choice. Maybe a setting to enable Name Service (NS) name, avatar or both. If users chose not to use a NS, there would also be less requests made.
I also like the idea of the badge if they choose to use the NS vs. using a different name for the account

No, it's not a problem. Since it's a reverse-lookup, if you have many names under 1 address it should only show the one that you "Make Primary" in the NS

src/routes/auth/connect.tsx
```diff
@@ -80,15 +90,6 @@ export function ConnectAuthRequestView() {
     gateway
   } = authRequest;
 
-  // wallet switcher open
```
pawanpaudel93, line 83: @7i7o Removing this is breaking the wallet api popups. Seems its removed only in`development` branch but not in `staging` and `production` branch.

nicholaswma: @pawanpaudel93 this was just merged in today - not sure what happened here, but I can revert these changes

pawanpaudel93: @nicholaswma we should bring those removed states back and test again with the connect api popup as it was affected.

nicholaswma: @pawanpaudel93 will resolve first thing tomorrow morning. we won't proceed with any releases until this is resolved

PR #659, opened 2025-02-12
feat: gateways settings

pawanpaudel93:
## Summary

This PR updates the `Wayfinder` to `Gateways` settings to match the mobile settings. Also, it updates the `findGateway` function to reflect this.

Related Jira Ticket: [ARC-1016](https://communitylabs.atlassian.net/browse/ARC-1016?atlOrigin=eyJpIjoiZWI3YjQ5MTRmMjkzNDAyNmI4MjBlYzdiODRiMjE2NjYiLCJwIjoiaiJ9)

[ARC-1016]: https://communitylabs.atlassian.net/browse/ARC-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #668, opened 2025-02-17
fix: hot fix, added random prop to gateway selector

nicholaswma:


7i7o:
@nicholaswma I think this one conflicts with Pawan's #659 
Check if you are not solving the same issue in both PRs

PR #669, opened 2025-02-17
chore: add embed ui

Jonathansoufer:


7i7o (approved):
tACK 👍 
This only adds the components but they are not connected

PR #677, opened 2025-02-19
fix: use a random fallback gateway

pawanpaudel93:
## Summary  

This PR improves the transaction retry mechanism in the send flow by selecting a random gateway as a fallback. Additionally, it enhances the error-handling experience by displaying a **gateway switch** button on the error toast, allowing users to manually switch gateways when a transaction fails.

**Jira Ticket:** [ARC-1112](https://communitylabs.atlassian.net/browse/ARC-1112)

## Screenshots  
image

## How to Test  
Since the issue might not be easily reproducible (depending on gateway availability), you can manually introduce an error in the code to:  
- Test the **gateway fallback** mechanism.  
- Verify that the **error toast** displays correctly with the gateway switch option.  

Let me know if you have any questions! 🚀 

[ARC-1112]: https://communitylabs.atlassian.net/browse/ARC-1112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/routes/popup/send/confirm.tsx
+++ b/src/routes/popup/send/confirm.tsx
@@ -343,11 +343,19 @@ export function ConfirmView({
           });
           navigate(`/send/completed/${res}?isAo=true`);
           setIsLoading(false);
+        } else {
+          throw new Error("Failed to send ao transfer");
         }
         return res;
       } catch (err) {
         console.log("err in ao", err);
-        throw err;
+        setIsLoading(false);
+        setToast({
+          type: "error",
+          content: browser.i18n.getMessage("failed_tx"),
+          duration: 2000
+        });
+        return;
       }
     }
     // Prepare transaction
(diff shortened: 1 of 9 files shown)


src/routes/popup/send/confirm.tsx
```diff
@@ -410,11 +444,7 @@ export function ConfirmView({
           console.log(e);
           setIsLoading(false);
           freeDecryptedWallet(keyfile);
-          setToast({
-            type: "error",
-            content: browser.i18n.getMessage("failed_tx"),
-            duration: 2000
-          });
+          showTransferError();
```
7i7o, line 453: Tested blocking the gateways, but if the gateways are not responding, we don't get to this part of the code

The `prepare` function already throws in line 209 of the code

In order to see the error message I need to block the gateways only when it's submitting the tx

Also, with gateways blocked from the start, it shows an error message and doesn't even reach the confirm screen of the flow
IMO most users with gateway issues won't see the error message

7i7o (commented):
Tested, but we need to re-think how we detect the gateway connectivity issues

7i7o (approved):
tACK 👍

PR #682, opened 2025-02-20
fix: enforce single instance

nicholaswma:


wander-embedded-sdk/src/wander-embedded.ts
```diff
@@ -47,6 +48,10 @@ export class WanderEmbedded {
   public pendingRequests: number = 0;
 
   constructor(options: WanderEmbeddedOptions = {}) {
+    if (WanderEmbedded.instance) {
+      throw new Error("WanderEmbedded instance already exists.");
```
Danziger, line 52: Maybe a getter to get the existing instance would be helpful.

Danziger (approved):

PR #683, opened 2025-02-20
feat: updated iframe to use shadow dom

nicholaswma:


wander-embedded-sdk/src/wander-embedded.ts
```diff
@@ -116,8 +120,8 @@ export class WanderEmbedded {
       this.backdropRef = elements.backdrop;
       this.iframeRef = elements.iframe;
 
-      document.body.appendChild(elements.backdrop);
-      document.body.appendChild(elements.iframe);
+      // document.body.appendChild(elements.backdrop);
```
Danziger, line 118: Either add all the elements to the body in `initializeComponents` or in `constructor`. That also applies to the `if (this.buttonRef) { ... }` block that was already there in the constructor.

Danziger (approved):

PR #684, opened 2025-02-20
feat: add minimize

nicholaswma:


Danziger (approved):

Danziger:
@nicholaswma I think you might need to add a minimize icon in some other places. If I recall correctly, there are a few different "header" components that are used in different views (e.g. request/permission views).

Also, the new authentication screens will also need that either added or connected to the logic to send the close message. You can sync with @Jonathansoufer on that.

nicholaswma:
@Danziger yea, I'll add it on all headers in the wallet then

Danziger (approved):

PR #685, opened 2025-02-20
fix: wallet switcher balances

pawanpaudel93:
## Summary  

This PR reintroduces the Redstone API as a fallback for AR balance fetching, ensuring data freshness by discarding prices older than 2 hours.


7i7o (approved):
tACK 👍

PR #696, opened 2025-02-24
fix: wallet view labels should not append .ar

kunstmusik:
NameServiceProfile data for ANS appends .ar to the profile name. Appending .ar here make it not work for either ArNS nor ANS.


nicholaswma (approved):
tACK 🚀

PR #701, opened 2025-02-26
fix: added wrapper for iframe contents

nicholaswma:


wander-embedded-sdk/src/components/iframe/wander-iframe.component.ts
```diff
@@ -234,10 +239,10 @@ export class WanderIframe {
     // style attribute must be reset to avoid conflicts with leftover properties from the previous layout
     if (resetLayout) {
```
Danziger: You can remove this, it was only needed before when the styles where added using a `style` attribute.

Danziger (approved):

wander-embedded-sdk/src/components/iframe/wander-iframe.component.ts
nicholaswma: @Danziger removed resetLayout

Danziger (approved):
utACK

PR #706, opened 2025-03-03
fix: api popups & embedded utils issue

pawanpaudel93:
## Summary  

This PR fixes the extension API popups by checking if `app` is `"wander"` instead of `ext` and correctly importing `EMBEDDED_PARENT_ORIGIN` in the background setup.

## How To Test
- Make sure extension api popups works as expected.


7i7o (approved):
tACK 👍

PR #710, opened 2025-03-05
fix: updated when pw is requested based on settings

nicholaswma:
https://communitylabs.atlassian.net/jira/software/c/projects/ARC/boards/17?selectedIssue=ARC-1152


7i7o (approved):
tACK 👍

PR #711, opened 2025-03-06
feat: application validation

pawanpaudel93:
## Summary

This PR adds the application validation logic.

**Note:** Please test this with the dashboard PR.

### DIFF
--- a/src/utils/embedded/embedded.utils.ts
+++ b/src/utils/embedded/embedded.utils.ts
@@ -4,7 +4,8 @@ import { IS_EMBEDDED_APP } from "~utils/embedded/embedded.constants";
 // Then, its tRPC client will be initialized with the following headers:
 // - authorization (getAuthTokenHeader / setAuthTokenHeader)
 // - x-device-nonce (getDeviceNonceHeader / setDeviceNonceHeader)
-// - x-api-key (getApiKeyHeader / setApiKeyHeader)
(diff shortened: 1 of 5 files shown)


wander-embedded-sdk/README.md
```diff
@@ -29,6 +29,8 @@
 
   useEffect(() => {
     const wanderInstance = new WanderEmbedded({
+      applicationId: "<applicationId>",
+      clientId: "<clientId>",
```
Danziger, line 32: Because client ID and application ID map 1-to-1, I think they should only provide the client ID (just like with Auth0, for example). With that, the verification endpoint can load the application already.

src/utils/embedded/embedded.utils.ts
```diff
@@ -76,15 +82,127 @@ const supabase = IS_EMBEDDED_APP
...
+      ">Invalid Wander Embedded configuration</h1>
+      <p style="
+        margin: 0;
+        font-size: 14px;
+        line-height: 1.5;
+        opacity: 0.9;
+      ">${errorMessage}</p>
+    `;
+
+    document.body.appendChild(overlay);
```
Danziger: Rather than appending it, I think it's better to replace the whole document, that's why I was using `document.write`. The idea is that this looks as close as possible to what users would get it validation was happening on the server side, meaning, if it fails, the app won't even load.

Danziger (approved):

PR #717, opened 2025-03-10
fix: authentication issues

pawanpaudel93:
## Summary

This PR addresses the authentication issues partially from [ARC-1150](https://communitylabs.atlassian.net/jira/software/c/projects/ARC/boards/17?selectedIssue=ARC-1150)

These issues are handled:
- When requests throw `UNAUTHORIZED`, the user is logged out.
- Handled oauth errors by redirecting to auth error page.
- Opened the authentication in a window popup rather than a tab.

Related PR: https://github.com/wanderwallet/embed-api/pull/22

[ARC-1150]: https://communitylabs.atlassian.net/browse/ARC-1150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/embedded/embedded.provider.tsx
```diff
@@ -735,32 +735,56 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
         if (url) {
-          // Redirect to Google's OAuth page
-          // Opening the URL on the current tab won't work when Embedded is loaded inside the iframe:
+          // Calculate center position for the popup
+          const width = 500;
+          const height = 600;
+          const left = window.screenX + (window.outerWidth - width) / 2;
+          const top = window.screenY + (window.outerHeight - height) / 2;
+
+          const popup = window.open(
```
Danziger, line 753: On what environment(s) did you try this? Just asking because there are some issues that will appear only when running the deployed app inside an iframe hosted on a different domain (like not being able to open the provider URLs within the iframe itself).

Otherwise, I'll be testing this once this PR is merged on the big one.

Might be worth reading: https://blog.huli.tw/2022/04/07/en/iframe-and-window-open/

package.json
```diff
@@ -81,7 +81,7 @@
     "copy-to-clipboard": "^3.3.2",
     "currency-symbol-map": "^5.1.0",
     "dayjs": "^1.11.6",
-    "embed-api": "https://github.com/wanderwallet/embed-api#e78ae041dfa1d2ca3d58dd2208f0bf6fb5a6266f",
+    "embed-api": "../embed-api",
```
Danziger: Just a note for myself: Revert once the backend PR is merged.

src/utils/embedded/embedded.utils.ts
```diff
@@ -72,21 +87,20 @@ const {
 // type TRPCClient = ReturnType<typeof createTRPCProxyClient<AppRouter>>;
 // const trpcVanilla = client as TRPCClient;
 
-const supabase = IS_EMBEDDED_APP
-  ? createSupabaseClient(
+function getSupabaseClient() {
```
Danziger, line 90: Now that you are here, I'd like to request one more change: Make this function async and provide a custom storage instance for supasbase that uses unpartitioned state (you can see a commented out draft for that in `src/iframe/plasmo-storage/plasmo-storage.mock.ts`).

src/utils/embedded/embedded.provider.tsx
pawanpaudel93: I have tried this in the development environment only.

Danziger:
Note for myself to test this later: Once we fix the unpartitioned state issue, all instances of Wander Embedded running on different sites will share the authentication state. However, after being redirected to the provider authentication page on a new tab or poup, the user is redirected back to the app to complete the authentication process. Most likely, for this to work, the user needs to be redirected to a page that also loads Wander Embedded inside an iframe. The reason for this is that the unpartitioned state storage is not the same as the storage used when the page is loaded on its own.

See https://stackoverflow.com/questions/71819128/supabase-auth-onauthstatechange-not-working-when-react-app-is-in-iframe

Danziger (approved):
utACK, but I'll be testing it on the other PR once I go back to it, as I'm also adding some changes on top.

PR #722, opened 2025-03-13
feat: unpartitioned storage for supabase auth & wallet

pawanpaudel93:
## Summary

This PR adds the support for unpartitioned localStorage which is used for storing the supabase auth keys so different websites using the embedded wallet can share the same auth & wallets.


src/utils/embedded/embedded.utils.ts
```diff
@@ -128,16 +130,36 @@ const {
...
+export async function getStorage() {
+  if (!storage) {
+    storage = new LocalStorage();
+    await storage.requestStorageAccess();
+  }
+
+  return storage;
+}
+
+export async function getSupabaseClient() {
```
Danziger, line 120: As we should just use the `getSupabaseClient` function, let's remove the `supabase` export below (collapsed in the PR). Let's keep the call to `getSupabseClient()` though to make sure we initialize this as soon as possible.

Danziger (approved):
utACK, do not worry about the comment, I can address it in the target branch once merged.

src/iframe/storage/plasmo-storage/plasmo-storage.mock.ts
```diff
@@ -0,0 +1,275 @@
...
+export class StorageMock extends PlasmoStorage implements StorageMockInterface {
+  private storage: EnhancedStorage;
+
+  constructor() {
+    super({ area: "session" });
+    this.storage = new EnhancedStorage({ area: "session" });
+
+    // This browser doesn't support the Storage Access API
+    // so let's just hope we have access!
+    if (!document.hasStorageAccess) return;
```
Danziger, line 60: If we come across this (or if users reject the unpartitioned state access request, maybe we should just use the regular `localStorage` but show an alert banner on the app letting them know their wallets will not be available on other sites automatically (without them manually importing or recovering them).

pawanpaudel93: Sure, will add the banner soon.

pawanpaudel93: Added here: https://github.com/wanderwallet/Wander/pull/732

PR #723, opened 2025-03-16
Enhance Encrypt Module to Support Custom Public Keys

ALLiDoizCode:
Currently, the `encrypt` module's asymmetric encryption implementation relies solely on the wallet's public key for encryption. This design delegates the responsibility of encrypting data with any alternative public key to the individual project, limiting flexibility.

This pull request updates the `encrypt` module to improve its functionality. With this change, the module will now check for a public key in the provided `options` object and use it for encryption if one is specified. If no public key is found in the `options` object, it will continue to default to the wallet's public key as before.

**Changes:**
- Modified the `encrypt` module to prioritize a public key from the `options` object when available.
- Maintained backward compatibility by falling back to the wallet's public key if no custom key is provided.

**Benefits:**
- Increases flexibility by allowing projects to specify custom public keys for encryption.
- Simplifies implementation for projects needing to encrypt data with keys other than the wallet's default.


Danziger:
@ALLiDoizCode Thanks for the PR, but I don't see why a wallet, whose main purpose is to manage the user's keys, should provide generic cryptography functionality (not related to the user's keys). I think that belongs to a utility library. Do you have a specific use-case in mind?

PR #726, opened 2025-03-19
feat: enhance storage api

pawanpaudel93:
## Summary

This PR adds storage methods for expiry, priority (for eviction) and access to unpartitioned state using Storage Access API.


src/iframe/storage/plasmo-storage/plasmo-storage.mock.ts
```diff
@@ -0,0 +1,275 @@
...
+export class StorageMock extends PlasmoStorage implements StorageMockInterface {
+  private storage: EnhancedStorage;
+
+  constructor() {
+    super({ area: "session" });
+    this.storage = new EnhancedStorage({ area: "session" });
+
+    // This browser doesn't support the Storage Access API
+    // so let's just hope we have access!
+    if (!document.hasStorageAccess) return;
```
Danziger, line 60: If we come across this (or if users reject the unpartitioned state access request, maybe we should just use the regular `localStorage` but show an alert banner on the app letting them know their wallets will not be available on other sites automatically (without them manually importing or recovering them).

PR #729, opened 2025-03-20
fix: update dynamic QR code to be scannable

pawanpaudel93:
## Summary  

This PR improves the QR code to ensure it is easily scannable on Android devices. Seems we were showing the QR code with white color and background with black color and Android was finding difficulty in scanning the QR.

| Before | After |  
|--------|-------|  
| image | image |

## How To Test
- Make sure its scannable in both iOS and Andorid.


7i7o (approved):
tACK 👍

PR #731, opened 2025-03-21
fix: display caching on txn detail page

nicholaswma:
@7i7o will update this PR when I've tested with keystone

I'll also look to clean up `[id].tsx` and `confirm.tsx`


7i7o (approved):
tACK 👍

PR #732, opened 2025-03-21
feat: add partitioned storage banner

pawanpaudel93:
## Summary

This PR shows the partitioned storage banner if access to unpartitioned storage is not available or denied.

Screenshot 2025-03-21 at 9 38 51 PM


src/iframe/storage/unpartitioned-storage/unpartitioned-storage.utils.ts
```diff
@@ -0,0 +1,5 @@
+export const PARTITIONED_STORAGE_BANNER_DISMISSAL_KEY =
+  "PARTITIONED_STORAGE_BANNER_DISMISSED";
+export const PARTITIONED_STORAGE_BANNER_MESSAGE =
```
Danziger: This needs to come from the translation files.

src/components/StoragePartitionedBanner.tsx
```diff
@@ -0,0 +1,210 @@
...
+import { useEmbedded } from "~utils/embedded/embedded.hooks";
+import {
+  PARTITIONED_STORAGE_BANNER_DISMISSAL_KEY,
+  PARTITIONED_STORAGE_BANNER_EVENT,
+  PARTITIONED_STORAGE_BANNER_MESSAGE
+} from "~iframe/storage/unpartitioned-storage/unpartitioned-storage.utils";
+
+// Styled components for the banner
+const BannerWrapper = styled(motion.div)`
+  position: fixed;
```
Danziger: I think it would be better if we make it non-dismissible, appears from the bottom but without `position: fixed` so that it doesn't cover the rest of the UI, and includes a button to re-request access.

The reason for this is that people might reject the permission without being aware of the consequences, and if they then dismiss the banner, they won't be able to get it back or to easily fix it (other than trying to reload the page).

Also, once the UI PR is merged, we need to make sure it appear in all screens, but I guess it's already the case.

Danziger:
@7i7o Once this branch is updated with the latest changes from `development` and the comments are addressed, it can be merged as well.

Danziger (approved):
utACK

PR #742, opened 2025-03-27
feat: encrypt recovery shares for secure storage and access

pawanpaudel93:
## Summary

This PR uses hybrid approach of RSA & AES encryption for storing recovery shares which can be used for faster re-download of recovery shares.


src/utils/embedded/embedded.provider.tsx
```diff
@@ -259,37 +264,122 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+
+    // Download the recovery file for the user
+    downloadRecoveryFile(walletAddress, recoveryFileData);
 
     // TODO: Make sure we use `freeDecryptedWallet` all over the place in the new code for Embedded:
     freeDecryptedWallet(jwk);
   }, [walletId, walletAddress]);
 
+  // Check if a wallet has a stored recovery share
+  const hasStoredRecoveryShare = useCallback(async (): Promise<boolean> => {
```
Danziger, line 339: I would remove `hasStoredRecoveryShare` and `retrieveStoredRecoveryShare` from the provider, as these functions are not used outside of it.

Danziger (approved):

src/utils/embedded/embedded.provider.tsx
pawanpaudel93: Sure, will remove it from the provider.

Danziger (approved):
utACK

PR #763, opened 2025-04-03
fix: Fix BE and iframe builds, but breaks SDK build.

Danziger:
In `feature/somorphic-message-utils`:

- ✅ Embed's build works.
- ❌ BE build doesn't work (it complains about `webextension-polyfill` being loaded in the browser. This is because `sendChunk()` was importing, but not using, it indirectly, when importing `extension-messaging.strategy.ts`.
- ❌ Both messaging strategies where imported in both builds (Embed and BE) even if only one of them was going to be used (based on an ENV variable in `messaging.utils.ts`.

Here:

- ✅ Embed's build works.
- ✅ BE's build works.
- ✅ Each build only imports the strategy that it needs.


package.json
```diff
@@ -7,8 +7,8 @@
   "packageManager": "yarn@1.22.18",
   "homepage": "https://www.wander.app",
   "scripts": {
-    "dev:chrome": "plasmo dev --verbose",
-    "build:chrome": "plasmo build --no-hoist",
+    "dev:chrome": "cross-env PARCEL_WORKER_BACKEND=process plasmo dev --verbose",
```
Danziger, line 10: See https://github.com/parcel-bundler/parcel/issues/10081#issuecomment-2736266818

PR #764, opened 2025-04-04
fix: ARC-1209 - Implement storage watchers and both mitt and custom events

Danziger:
This PR implements storage watchers and both mitt and custom events, which fixes the issue with the connection status indicator on the `WanderButton` not updating when the wallet got disconnected from the app, as well as fixing the previously not working events.

It also adds proper functionality to the SDK to listen to the various types of events in order to update the UI accordingly and to inject the wallet API and trigger the wallet ready event when appropriate.

Closes:
- https://communitylabs.atlassian.net/browse/ARC-1203 - SDK auth button missing loading state
- https://communitylabs.atlassian.net/browse/ARC-1209 - SDK connection indicator not updating properly
- https://communitylabs.atlassian.net/browse/ARC-1257 - SDK theming options not working properly
- https://communitylabs.atlassian.net/browse/ARC-1258 - SDK popup/modal resizing not working properly
- https://communitylabs.atlassian.net/browse/ARC-1260 - Connect Sign Out button missing
- https://communitylabs.atlassian.net/browse/ARC-1261 - Connect auth provider buttons are disabled
- https://communitylabs.atlassian.net/browse/ARC-1262 - Connect Wander "authentication" button not working

Related to https://github.com/Othent/KMS-test-repo/pull/19.

Detailed breakdown:

- [x] Add sign out button under actions
- [x] Enable all auth providers.
- [x] Enable Wander "auth option" and implement it.
  - [x] Hide it if there's not BE installed or if it's not Wander.
- [x] Include name, picture... in `UserDetails` / `getUserDetailsFromSupabaseUser()`.
- [x] Remove `embedded_connect` and `embedded_disconnect` and instead use the existing wallet events.
- [x] Dedupe events sent from/by the SDK.
- [x] `arweaveWalletLoaded` event updates:
  - [x] `arweaveWalletLoaded` now returns the permissions for both BE and Connect.
  - [x] For Connect, `arweaveWalletLoaded` once the wallet is ready to be used.
  - [x] For BE, `arweaveWalletLoaded` is now triggered making sure no listener can never receive the event (if the dApp starts listening too late).
- [ ] ~~`connect` event updates:~~ Changes not implemented in the end.
  - [ ] ~~For Connect, the `connect` event now returns the permissions.~~
  - [ ] ~~For BE, the `connect` should also return the permissions.~~
- [x] Add new SDK options / URL params:
  - [x] `theme = "system" | "light" | "dark"`
  - [x] `hideBE = boolean`
- [x]  Fix auth buttons being re-enabled too early, before the redirect to the dashboard happens.
- [x] Fix back up later + don't ask again redirect.


src/utils/messaging/strategies/iframe/iframe-messaging.strategy.ts
```diff
@@ -20,8 +20,8 @@ const messageHandlersByMessageID: Partial<
   Record<MessageID, Set<OnMessageCallback<MessageID>>>
 > = {};
 
-let targetIframe: HTMLIFrameElement = null;
-let targetOrigin = "";
+let embedWIndow: Window = null;
```
Danziger: Typo.

assets/popup.css
```diff
@@ -72,13 +72,21 @@
   }
 }
 
-body {
-  margin: 0;
-  padding: 0;
+html[data-theme] {
```
Danziger, line 75: This applies to Connect only, while the one below applies to BE only.

Not the nicest, but we can reorganize these shared files after the monorepo is setup.

src/api/foreground/foreground-setup-embedded-wallet-sdk.ts
```diff
@@ -132,29 +132,4 @@ export function setupEmbeddedWalletSDK(
 
   // @ts-expect-error
   window.arweaveWallet = walletAPI;
-
-  // at the end of the injected script,
-  // we dispatch the wallet loaded event
-  dispatchEvent(new CustomEvent("arweaveWalletLoaded", { detail: {} }));
```
Danziger, line 138: For Connect, these are triggered by the SDK after 1) it's instantiated, 2) the wallet API is injected into the page and 3) the authentication process completes and the wallet is ready to be used.

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -177,4 +167,66 @@ export function setupWalletSDK(
...
+  async function dispatchArweaveWalletLoaded() {
+    if (!window.arweaveWallet) return;
+
+    const permissions = await window.arweaveWallet
+      .getPermissions()
+      .catch(() => []);
+
+    // Note that for Wander Connect we just need to dispatch this once, no need to subscribe to the window load event to re-dispatch it:
+    dispatchEvent(
+      new CustomEvent("arweaveWalletLoaded", {
```
Danziger, line 183: The `arweaveWalletLoaded` now returns the current dApp permissions.

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -177,4 +167,66 @@ export function setupWalletSDK(
...
+        detail: {
+          permissions
+        }
+      })
+    );
+
+    if (permissions.length > 0) {
```
Danziger, line 190: @7i7o I think it makes sense that when a dApp is connected, it receives the `arweaveWalletLoaded` as well as all the other events, as that gives it the current state of the wallet.

However, with the current implementation, if I add a new listener to `arweaveWalletLoaded` way later after the page has loaded, all these other events are dispatched again. Maybe instead we could just include all this in the `arweaveWalletLoaded` details (permissions, activeAddress, addresses, gateway, ...).

src/api/foreground/foreground-setup-wallet-sdk.ts
```diff
@@ -177,4 +167,66 @@ export function setupWalletSDK(
...
+        window.arweaveWallet.getActiveAddress().catch(() => ""),
+        window.arweaveWallet.getAllAddresses().catch(() => [])
+      ]);
+
+      events.emit("activeAddress", activeAddress);
+      events.emit("addresses", addresses);
+    }
+  }
+
+  // Not sure there's a point in this, as the dApp will never be able to listen for it, but I'm maintaining the same structure we had before:
```
Danziger, line 203: @7i7o See my comments here and below. Maybe this line is no longer needed.

src/api/modules/connect/connect.background.ts
```diff
@@ -24,9 +31,31 @@ const background: BackgroundModuleFunction<void> = async (
 
   if (gateway) isGateway(gateway);
 
-  // Note we are not checking if there are any wallets added anymore, as `requestUserAuthorization` (actually
-  // `createAuthPopup`) do that automatically now and will open the Welcome page and throw a "No wallets" added error if
-  // there are no wallets.
+  // While `requestUserAuthorization` (actually `createAuthPopup`) already does this check below, opening the Welcome
```
Danziger, line 34: This check probably needs to go into `disconnect` and `getArweaveConfig`, as those work before the wallet is ready. `getPermissions` too but that one seems to work before the wallet is ready for BE too.

src/api/foreground/foreground-setup-wallet-sdk.ts
7i7o: I think this could be an improvement on how things are handled rn for events. Yet we would need to do both to not break backwards compatibility
On top of that, we have permissions to access adressess and activeAddress, so we would need to be compliant to that as well when we send all these data and respect the permissions given by the user

src/api/modules/connect/connect.background.ts
Danziger: https://communitylabs.atlassian.net/browse/ARC-1275

src/api/foreground/foreground-setup-wallet-sdk.ts
Danziger: https://communitylabs.atlassian.net/browse/ARC-1276

7i7o (approved):
tACK 👍

Danziger:
@7i7o regarding your last comment:

> I think this could be an improvement on how things are handled rn for events. Yet we would need to do both to not break backwards compatibility
> On top of that, we have permissions to access adressess and activeAddress, so we would need to be compliant to that as well when we send all these data and respect the permissions given by the user

I'm inclining towards not emitting the Mitt events (so removing that change), and just updating `arweaveWalletLoaded` to include `activeAddress`, `addresses`, `gateway`, etc. in `e.detail`, just like I added `permissions`.

Regarding making sure we take the current permissions into account, that should be covered already as we are simply doing wallet API calls, not accessing the data directly, so if the permission is not there for a specific method, it will throw an error (thus the `catch` blocks).

PR #767, opened 2025-04-10
fix: memory leak issue

pawanpaudel93:
## Summary

This PR does the following:

- Fixes the circular dependency with the code.
- Fixes the memory leak caused by the `node-forge` usage.


7i7o (approved):
tACK 👍

PR #772, opened 2025-04-21
fix: shares warning & errors

pawanpaudel93:
## Summary

This PR updates the code to throw shares missing error & different warning in console when device share is missing.

Related: https://communitylabs.atlassian.net/browse/ARC-1196


Danziger:
@pawanpaudel93 I think this just changes the error message users will see, but we need to find out why `generateShareHashAndPrivateKey` is being called with `share = null`, as it might be just an initialization issue.

Danziger:
Closing, as the fix has been included in https://github.com/wanderwallet/Wander/pull/830

Danziger (approved):

PR #773, opened 2025-04-21
fix(embedded): get app logo if not provided and improve transaction message

pawanpaudel93:
## Summary

This PR makes sure we try to get the app logo if not provided in the connect API. Also, it updates the logo style.

## Screenshots
| Connect | Sign |
|--------|--------|
|  |  |

## How to Test
1. Build the wallet api: 
    ```
    yarn build:wallet-api
    ```
2. Build the wander embedded SDK:
      ```
      cd wander-embedded-sdk && pnpm build
      ```
3. Install the built SDK and test it

### DIFF
--- a/src/api/modules/connect/connect.foreground.ts
+++ b/src/api/modules/connect/connect.foreground.ts
@@ -3,6 +3,8 @@ import type { AppInfo } from "~applications/application";
 import type { ModuleFunction } from "~api/module";
 import { type Gateway } from "~gateways/gateway";
 import { getAppURL } from "~utils/format";
+import { IS_EMBEDDED_APP } from "~utils/embedded/embedded.constants";
+import { getAppLogo } from "~utils/embedded/embedded.utils";
 
 const foreground: ModuleFunction<any[]> = async (
   permissions: PermissionType[],
@@ -22,6 +24,10 @@ const foreground: ModuleFunction<any[]> = async (
 
     // use site title if it is < than 11 chars
     appInfo.name = siteTitle.length < 11 ? siteTitle : tabURL;
+  }
+
+  if (IS_EMBEDDED_APP && !appInfo.logo) {
+    appInfo.logo = getAppLogo();
   }
 
   return [permissions, appInfo, gateway];
--- a/src/api/modules/connect/connect.foreground.ts
+++ b/src/api/modules/connect/connect.foreground.ts
@@ -4,7 +4,7 @@ import type { ModuleFunction } from "~api/module";
 import { type Gateway } from "~gateways/gateway";
 import { getAppURL } from "~utils/format";
 import { IS_EMBEDDED_APP } from "~utils/embedded/embedded.constants";
-import { getAppLogo } from "~utils/embedded/embedded.utils";
+import { getAppLogo } from "~utils/embedded/utils/logo/logo.utils";
 
 const foreground: ModuleFunction<any[]> = async (
   permissions: PermissionType[],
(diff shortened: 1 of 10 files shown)


src/utils/embedded/embedded.utils.ts
```diff
@@ -205,6 +205,50 @@ async function insecurelyValidateApplication() {
...
+      "link[rel='icon'][type='image/svg+xml']",
+      "link[rel='icon'][sizes='32x32']",
+      "link[rel='icon'][sizes='48x48']",
+      "link[rel='icon'][sizes='96x96']",
+      "link[rel='icon']",
+      "link[rel='shortcut icon']",
+      "link[rel='apple-touch-icon'][sizes='180x180']",
+      "link[rel='apple-touch-icon']",
+      "link[rel='apple-touch-icon-precomposed']",
+      "meta[property='og:image']",
```
Danziger: `og:image` is usually a rather large image meant to be displayed as a preview when sharing a link, so I would avoid using it as icon.

src/utils/embedded/embedded.utils.ts
```diff
@@ -205,6 +205,50 @@ async function insecurelyValidateApplication() {
...
+  } catch {
+    return null;
+  }
+};
+
+export function getAppLogo(): string | null {
+  try {
+    const faviconSelectors = [
+      "link[rel='icon'][type='image/svg+xml']",
+      "link[rel='icon'][sizes='32x32']",
```
Danziger: Maybe it's better to remove the selectors with `sizes` and instead change `document.querySelector` to `document.querySelectorAll`, then sort by the sizes attribute (when available).

src/utils/embedded/embedded.utils.ts
```diff
@@ -205,6 +205,50 @@ async function insecurelyValidateApplication() {
...
+      "link[rel='icon'][sizes='48x48']",
+      "link[rel='icon'][sizes='96x96']",
+      "link[rel='icon']",
+      "link[rel='shortcut icon']",
```
Danziger: I would move this one just above `img[alt*='logo']`, as it tends to be a 16x16 favicon.

src/utils/embedded/embedded.utils.ts
```diff
@@ -205,6 +205,50 @@ async function insecurelyValidateApplication() {
...
+  }
+  try {
+    return new URL(url, window.location.origin).href;
+  } catch {
+    return null;
+  }
+};
+
+export function getAppLogo(): string | null {
+  try {
```
Danziger: You could consider getting the icon from the manifest before trying to get it from the page. See https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest

src/api/modules/connect/connect.foreground.ts
```diff
@@ -24,6 +26,10 @@ const foreground: ModuleFunction<any[]> = async (
     appInfo.name = siteTitle.length < 11 ? siteTitle : tabURL;
   }
 
+  if (IS_EMBEDDED_APP && !appInfo.logo) {
```
Danziger, line 29: Any reason to use this in Embed but not in the BE? I would probably add this functionality to both if there's no issue with that.

pawanpaudel93: BE already uses the tab favicon if no logo is provided in the connect request:

https://github.com/wanderwallet/Wander/blob/5e69fe7920c91e94de613ab245e5609d225ceb4b/src/api/background/handlers/message/api-call-message/api-call-message.handler.ts#L110

https://github.com/wanderwallet/Wander/blob/5e69fe7920c91e94de613ab245e5609d225ceb4b/src/api/modules/connect/connect.background.ts#L50

Danziger (approved):

PR #776, opened 2025-04-22
fix: endless loading after unlock

pawanpaudel93:
## Summary

This fixes an issue where the extension gets stuck in a loading state after unlocking from a message request without any further popups.

## How to Test
- See Jira ticket: https://communitylabs.atlassian.net/browse/ARC-1219


7i7o (approved):
tACK 👍

Danziger (approved):
This will be re-fixed inside the AuthProvider, rather than the router, in https://github.com/wanderwallet/Wander/pull/779, but we'll merge it as it is to release a new version ASAP.

PR #779, opened 2025-04-24
fix: ARC-1219: Fix endless loading in provider

Danziger:
In the end, this PR only includes minor type, constant and props renames and:

- The addition of 1 new TS type.
- A conditional to only call `initializeDeviceNonce()` in Connect.

The original intention of this PR was to take the fix merged in https://github.com/wanderwallet/Wander/pull/776 and move its logic from the router to the `AuthProvider`. However, it looks like that was already addressed in https://github.com/wanderwallet/Wander/pull/830.


7i7o (approved):
tACK 👍

PR #793, opened 2025-04-30
feat: passkey integration

matteyu:
Blocking PR: https://github.com/wanderwallet/embed-api/pull/4

Client integration for passkey.  New integration for sign-in and registration for passkey.


src/utils/embedded/embedded.provider.tsx
```diff
@@ -1101,48 +1118,66 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
         if (activatedWallet) {
           const { id: walletId, address: walletAddress } = activatedWallet;
 
-          const jwk = await WalletUtils.generateWalletJWKFromShares(
-            walletAddress,
-            [activatedWallet.authShare, activatedWallet.deviceShare]
-          );
+          // Check if device share exists for passkey auth
+          const deviceShares = await WalletUtils.getDeviceSharesForUser(userId);
+          const hasDeviceShare = !!deviceShares[walletId];
```
Danziger: Inside this `if`, a wallet has already been activated, so this check should always pass. In any case, I'm not sure why these checks related to passkeys are merged together with the wallet activation flow, which should be authentication-method agnostic. That is, whatever logic needs to run around authentication, should happen before this block. Once we are here, authentication should be good.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1080,6 +1086,17 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
 
       let authStatus = "noAuth" as AuthStatus;
 
+      // Check for passkey authentication (should be flagged in localStorage)
+      const isPasskeyAuth = embeddedContextAuth.authProviderType === "PASSKEYS";
+      const isCustomAuth = localStorage.getItem("isCustomAuth") === "true";
```
Danziger: Why is this needed? Would passkeys still work if the user cleans the browser data?

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1214,6 +1249,69 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+          userId: customUserId,
+          deviceNonce: localStorage.getItem("deviceNonce") || undefined,
+          createdAt: new Date(),
+          updatedAt: new Date(),
+          ip: "127.0.0.1",
+          userAgent: navigator.userAgent
+        };
+
+        // Create a custom auth token for API requests
+        const deviceNonce =
```
Danziger: You should use `getDeviceNonce` here, which takes care of loading/initializing/storing it.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1214,6 +1249,69 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+            user_id: customUserId,
+            session_id: customSessionId,
+            device_nonce: deviceNonce,
+            exp: Math.floor(Date.now() / 1000) + 60 * 60, // 1 hour expiry
+            iat: Math.floor(Date.now() / 1000)
+          })
+        );
+
+        // Set the auth token header for API requests
+        setAuthTokenHeader(temporaryAuthToken);
```
Danziger: This should not work, as this token is not signed.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1275,18 +1373,106 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+            updatedAt: new Date(),
+            ip: "127.0.0.1",
+            userAgent: navigator.userAgent
+          };
+
+          // Clear the flag
+          localStorage.removeItem("needsWalletActivation");
+
+          // Important: Generate and set an authentication token for API requests
+          const deviceNonce =
```
Danziger: This should be done with `getDeviceNonce()`.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1275,18 +1373,106 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+            localStorage.getItem("deviceNonce") || crypto.randomUUID();
+
+          // Store this nonce for future requests
+          localStorage.setItem("deviceNonce", deviceNonce);
+
+          // Create a custom auth token to use for API requests
```
Danziger: Ok, this answer the question above about the tokens not being valid. The tokens must be signed by the server, as otherwise anyone could use this custom auth header to steal someone else's session/account.

Danziger:
A few lines salvaged here: https://github.com/wanderwallet/Wander/pull/982

PR #794, opened 2025-05-01
fix: Prevent API initiated popup auto cancel

pawanpaudel93:
## Summary

This PR adds the request arrival window of 1 seconds so that when the popup opens it doesn't auto close.

Related Jira Ticket: [ARC-1265](https://communitylabs.atlassian.net/browse/ARC-1265)

[ARC-1265]: https://communitylabs.atlassian.net/browse/ARC-1265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/auth/auth.provider.tsx
+++ b/src/utils/auth/auth.provider.tsx
@@ -2,12 +2,14 @@ import {
   createContext,
   useCallback,
   useEffect,
+  useRef,
   useState,
   type PropsWithChildren
 } from "react";
 import {
   AUTH_POPUP_CLOSING_DELAY_MS,
   AUTH_POPUP_REQUEST_WAIT_MS,
+  AUTH_POPUP_REQUEST_ARRIVAL_WINDOW_MS,
   AUTH_POPUP_UNLOCK_REQUEST_TTL_MS,
   ERR_MSG_USER_CANCELLED_AUTH
 } from "~utils/auth/auth.constants";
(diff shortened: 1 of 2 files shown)


src/utils/auth/auth.provider.tsx
```diff
@@ -500,16 +505,22 @@ export function AuthRequestsProvider({ children }: PropsWithChildren) {
...
           clearCloseAuthPopupTimeout = closeAuthPopup(
             AUTH_POPUP_UNLOCK_REQUEST_TTL_MS
           );
-        } else if (!hasAuthRequests) {
+        } else if (!hasAuthRequests && !isInRequestArrivalWindow) {
           // Once the wallet is unlocked, we close the popup if an AuthRequest doesn't arrive in less than
-          // `AUTH_POPUP_REQUEST_WAIT_MS` (1 second):
+          // `AUTH_POPUP_REQUEST_WAIT_MS` (1 second) but only if the component is not in the initial request arrival window:
           clearCloseAuthPopupTimeout = closeAuthPopup(
             AUTH_POPUP_REQUEST_WAIT_MS
```
Danziger, line 525: I think the same result can be achieved by simply increasing `AUTH_POPUP_REQUEST_WAIT_MS` to 2 seconds, as this `useEffect` already runs when the component mounts.

Also, not sure this works as expected, because if the `useEffect` runs too early (so `isInRequestArrivalWindow = false`), then the popup won't close even if a request never arrives, because the `useEffect` won't re-run after `AUTH_POPUP_REQUEST_ARRIVAL_WINDOW_MS` has passed.

pawanpaudel93: The same issue occurs when updating AUTH_POPUP_REQUEST_WAIT_MS to 2 seconds also.

Danziger (approved):

PR #798, opened 2025-05-02
Add action to store builds as artifacts

7i7o:
Adds an action to manually create builds from a branch and store it as an artifact in GH


pawanpaudel93 (approved):
utACK

Issue #725, opened 2025-03-19
yarn.lock upgrade / release process?


psm384:
More of a question, but perhaps an issue.

What's the release process with respect to the "yarn.lock" file? The README does not encourage users to run "yarn upgrade".

For example, if I pull 1.24.6, then pre-upgrade an audit analysis yields:

```
=== SEVERITY BREAKDOWN ===
Severity   | Raw Count  | Unique Count
----------------------------------------
Critical   | 9          | 1
High       | 4          | 2
Moderate   | 29         | 13
Low        | 46         | 6

=== UNIQUE VULNERABILITIES ===
ID         | Severity   | Module               | Created Date         | Title
---------------------------------------------------------------------------------------------------------
1102901    | critical   | elliptic             | 2025-02-12           | Elliptic's private key extraction in ...
1102472    | high       | axios                | 2025-03-07           | axios Requests Vulnerable To Possible...
1098393    | high       | ws                   | 2024-06-17           | ws affected by a DoS when handling a ..
```

and if i run an upgrade i get:

```
=== SEVERITY BREAKDOWN ===
Severity   | Raw Count  | Unique Count
----------------------------------------
Critical   | 0          | 0
High       | 0          | 0
Moderate   | 4          | 3
Low        | 1          | 1

```

eg zero critical or high issues, but if i didn't know to do the upgrade, i'd have a few serious issues left.

Danziger:
@psm384 Thanks for the heads-up. We are upgrading all dependencies now but we'll soon migrate to a monorepo. Once we do that, we'll drastically improve the whole CI/CD pipeline.

PR #799, opened 2025-05-05
Update dependencies

Danziger:
Update break down:

- Upgrade Husky (minor changes).
- Upgrade `shamir-secret-sharing` after their [last audit](https://github.com/Zellic/publications/blob/master/Privy_Shamir_Secret_Sharing_-_Zellic_Audit_Report.pdf) (no changes).
-  Upgrade `@keystonehq/arweave-keyring` and `@keystonehq/bc-ur-registry-arweave` from `0.1.1-alpha.0` to `^0.5.3` (requires testing).
- Upgrade Prettier. I'll make a follow-up PR to format all the code and update a couple of options.
- Upgrade `check-password-strength` to v3, which:

  > v3: allow all symbols by default (any character except the 26 latin lowercase, uppercase letters and 10 digits) & set the default min length to 12 instead of 10

- Upgrade `ar-gql` to v3 (we only use types).

- Upgrade TypeScript to v5.

- `@semantic-release/github` and `semantic-release` not updated.
- React and `styled-components` not updated (until we bring all the components into the monorepo).

Other changes:

- Remove 3 remaining usages of UUIDv4 in favor of nanoid.
- Update one remaining import from `warp-arbundles`.
- Disable analytics for Wander Connect (temporarily until we update that to).
- Prevent `Balance` from sending `embedded_balance` messages in the BE build.
- Close https://github.com/wanderwallet/Wander/issues/725.


.husky/pre-commit
```diff
@@ -1,5 +1 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
 yarn fmt
-git add -A
```
Danziger, line 5: This would include all Prettier-formatte files when you commit (good) but also include files you didn't add to that commit (bad). I'll change it in a new PR to use `lint-staged` or something similar.

7i7o: Looking forward to this change
I'm not a big fan of the `git add -A` on prettier pre-commits 🤣

7i7o (approved):
utACK 👍

PR #809, opened 2025-05-07
Minor UI fixes

Danziger:
Fix isAnchor property being added as an HTML attribute, and fix redirect after wallet backup not working


7i7o (approved):
utACK 👍

PR #816, opened 2025-05-08
chore: use --frozen-lockfile option for yarn install in workflow files

pawanpaudel93:
## Summary

This PR updates the workflow files to use `--frozen-lockfile` with dependency install.


7i7o (approved):
utACK 👍

PR #824, opened 2025-05-15
feat: Use cross-site cookies on Firefox to achieve cross-site authentication

pawanpaudel93:
This PR adds support for cookie storage of `deviceNonce` and Supabase's auth token, but is currently disabled in all browsers until further testing is done on that front.

**Previous PR description:**

This PR improves the unpartitioned storage handling of `deviceNonce` and Supabase's auth token. See [ARC-1286](https://communitylabs.atlassian.net/browse/ARC-1286).

Now, those two values will be stored in cookies instead of `localStorage` on browser that do not support unpartitioned state or didn't allow it (except in `localhost`). The following has been tested in the preview environment:

- Safari: According to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore#browser_compatibility), Safari doesn't support the partitioned option in `cookieStorage`. Therefore, it will still requires user to re-authenticate on each tab, so it doesn't make sense to use cookies there as that doesn't solve anything.

- Firefox: After authenticating on one site and opening Wander Connect on a different site, it authenticates automatically but requires recovering the wallet. After that, reloading both tabs keeps the user authenticated and the wallet activated.

  However, each of them has a different device nonce, most likely because it is created before the cross-site cookies permission is requested.

- Brave: After authenticating on one site (without giving access to embed content) and opening Wander Connect on a different site, it doesn't authenticate automatically. The check in `shouldStoreInCookies()` should be updated to work in that case, but that requires access to third-party cookies. For now, this will be kept like this, given it would be rare to see users not giving unpartitioned state access.

- Edge: Same.

- Chrome: Same.

See https://communitylabs.atlassian.net/jira/software/c/projects/ARC/boards/17?selectedIssue=ARC-1511.

[ARC-1286]: https://communitylabs.atlassian.net/browse/ARC-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -42,4 +89,536 @@ export class LocalStorage extends EnhancedStorage {
...
   }
+
+  /**
+   * Determine the primary storage strategy based on access state
+   * @returns Whether to use cookies as primary storage
+   */
+  /*
+  private shouldUseCookiesAsPrimary(): boolean {
+    // Only use cookies as primary when we have cookie access but not full access
+    return this.storage.hasCookieAccess() && !this.storage.hasFullAccess();
```
Danziger: @pawanpaudel93 Any reason why this was needed? We cannot save most stuff using cookies as then that data will be transferred to the server, potentially exposing device shares, for example.

Therefore, unless there's a reason for this, I've removed this function and changed the logic to only use localStorage or cookies (but not both), and cookies only in a very specific scenario and only for device nonce and Supabase auth token (see TSDoc comment below).

src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -42,4 +89,536 @@ export class LocalStorage extends EnhancedStorage {
...
+        // Fallback to js-cookie
+        Cookies.set(key, value, cookieOptions);
+        const result = Cookies.get(key);
+        return result === value;
+      }
+    } catch (e) {
+      if (LocalStorage.DEBUG_MODE) console.warn(`Failed to set cookie for key '${key}'`, e);
+
+      /*
+      try {
```
Danziger, line 172: Is there any reason for so many fallbacks? There's already the `if (cookieStore)` check above, but we still try to use `Cookies` in the `catch` block, and then also try with `localStorage`. Same question applies to the other methods that have this same pattern.

src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -42,4 +89,536 @@ export class LocalStorage extends EnhancedStorage {
...
+      if (LocalStorage.DEBUG_MODE) console.error("Error in removeChunkedCookies", e);
+    }
+  }
+
+  /**
+   * Set an item in storage, with appropriate backup strategies
+   */
+  async setItem<T>(key: string, value: T, useRaw = false): Promise<void> {
+    if (!key) {
+      // TODO: Wouldn't this throw an error already? Otherwise, it should.
```
Danziger: @pawanpaudel93 Any reason why you added a `return` instead of allowing the error to be thrown (I guess one will be thrown in some of the lines below if no key is passed).

src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -42,4 +89,536 @@ export class LocalStorage extends EnhancedStorage {
...
+   * Set a cookie with chunking if needed
+   */
+  private async setCookieWithChunkingIfNeeded(key: string, value: string): Promise<boolean> {
+    if (!value) {
+      if (LocalStorage.DEBUG_MODE) console.warn(`Attempted to set empty value for key '${key}'`);
+      return false;
+    }
+
+    // Check if value is too large for a single cookie
+    if (value.length > LocalStorage.MAX_COOKIE_CHUNK_SIZE) {
```
Danziger, line 291: I think some browsers do allow for way larger cookies, so trying to set it in a `try..catch` block is probably a better approach.

src/iframe/storage/unpartitioned-storage/local-storage.ts
pawanpaudel93: I think I used cookieStore but it could throw error so added a fallback with `js-cookie` incase if it failed in the catch block. I think regarding localStorage, i was trying to sync the localStorage and cookies so they have the same value if in a case where the storage can switch from cookie to localStorage if permissions was granted later.

src/iframe/storage/unpartitioned-storage/local-storage.ts
pawanpaudel93: Yes, I think its browser dependent and some browser was letting to store in a single key-value pair and some needed to store in two key-value pairs so I added chunked cookie storing. I did some research and 4KB was the max value size but browsers were storing the same value differently with metadata so i choosed the max value like that.

src/iframe/storage/unpartitioned-storage/local-storage.ts
pawanpaudel93: Yes, I think this can be removed as I don't think I added this for some purpose.

src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -1,18 +1,77 @@
...
+      partitioned: false,
+    };
+  }
+
+  private constructor() {
+    this.storage = new EnhancedStorage({ area: "local" });
+
+    this.isHttps = typeof location !== "undefined" && location.protocol === "https:";
+
+    // Safari doesn't support the partitioned option in cookieStorage. See https://developer.mozilla.org/en-US/docs/Web/API/CookieStore#browser_compatibility
```
Danziger, line 50: @pawanpaudel93 If you remove `!browserInfo.isSafari` from the line below, does this work for you on Safari (e.g. if you authenticate on WC on site A, site B also authenticates automatically).

pawanpaudel93: Let me check that and comment.

pawanpaudel93: @Danziger In Firefox, if you stop blocking cross-site cookies (either from settings or by disabling Enhanced Tracking Protection via the shield), we don’t need to store the Supabase and device nonce in cookies. Using localStorage works fine — I tested on `https://connect-dev.wander.app`, signed in on one tab, and the other tab was also signed in.

image

image

pawanpaudel93: @Danziger I tested this on Safari as you suggested by removing `!browserInfo.isSafari`, but it doesn’t work as unless “Prevent cross-site tracking” is disabled, Safari won’t store the cookies.

This behaves the same way as Firefox — if you disable “Prevent cross-site tracking”, it works fine: storing device nonce and supabase auth in cookies aren’t needed, and localStorage works fine, and the signed-in wallet loads correctly across tabs.

I tested with: https://connect-dev.wander.app

image

Danziger (approved):
tACK, with most of the new functionality disabled as stated.

PR #826, opened 2025-05-16
refactor: improve recover account import

pawanpaudel93:
## Summary

This PR ensures that the keyfile or seed phrase is only required once during the account recovery flow, simplifying the user experience.


7i7o (approved):
tACK 👍

PR #807, opened 2025-05-07
feature/ARC-1256: Loading states for auth/account screens

Danziger:
This major revamp includes:

**Share restore:**
- [x] Create specific `auth-restore-shares-keyfile.view.tsx` and `auth-restore-shares-seedphrase.view.tsx` and clearly differentiate them in the copy from `auth-import-keyfile.view.tsx` and `auth-import-seedphrase.view.tsx`. Also, add a wallet address confirmation step to them.
- [x] Make `auth-restore-shares-recovery-file.view.tsx` and `auth-restore-shares-keyfile.view.tsx` both work with the other file type (in case users get confused).
- [x] Fix back button in the recovery screens not doing anything (it should sign out the user and show the authentication selection screen again). Closes https://communitylabs.atlassian.net/browse/ARC-1274.

**Email authentication:**
- [x] Remove `authEmail` form `EmbeddedProvider` and instead use search params.

**`Card` component:**
- [x] Implement and add to all views a loading state. Closes https://communitylabs.atlassian.net/browse/ARC-1256.
- [x] Create `OnboardingCard` molecule to avoid repeating `Card` components, which can use a `<form>` as root element when needed.

**`Button` component:**
- [x] Properly type `href` (as `WanderRoutePath`) and remove `href` values starting with `#`.

**Input components and accessibility:**
- [x] Make the `Tabs` component work for any arbitrary number of tabs, and style it like other inputs.
- [x] Make sure all onboarding screens can be navigated pressing Tab. Closes https://communitylabs.atlassian.net/browse/ARC-1287
- [x] Remove dead (non-clickable / non interactable) zones form inputs.
- [x] Create a new `PasswordInput`.
- [x] Remove default Chrome background color when inputs are auto-filled.

**Misc.:**
- [x] Simplify auth/onboarding screen layouts: Many `Box` and `Row` components have been removed, as well as many `padding` / `margin` properties. Layout happens mostly with `display: flex` + `gap: X` on the parent container.
- [x] Seed input width doesn't match the rest of the elements in its same view / container.
- [x] Rewire the back/close button logic so that the close one always closes and the back one always goes to the previous (by hierarchy) view.
- [x] SDK button stays in loading mode when closing the auth provider popup window. Coses https://communitylabs.atlassian.net/browse/ARC-1305.
- [x] Use `<form>` + `onSubmit` instead of checking of listening for `keydown` to manually "submit" the data.
- [x] Fix routing issue when using the back button in the More auth methods view.

[...]

PR #830, opened 2025-05-18
ARC-1307: Auth request screens and logic improvements & fixes

Danziger:
- [x] Fix Connect requests getting stuck if received while the backup reminder is being shown. Now, the Connect request view won't be displayed until the backup reminder (and the account confirmation) is handled. Closes https://communitylabs.atlassian.net/browse/ARC-1307.
- [x] Update auth request views to not close the modal until the last one is handled. Closes https://communitylabs.atlassian.net/browse/ARC-1293.
- [x] Update auth request views to react to keys consistently (Y/Enter = Confirm, N/Delete = Reject, Close icon minimizes (without rejecting)).
- [x] Clear tokens and apps from storage when signing in with a different user. Fixes https://communitylabs.atlassian.net/browse/ARC-1301 and https://communitylabs.atlassian.net/browse/ARC-1300 at the cost of providing a worse experience for users that share a device, and doesn't add the multi-wallet support that would also be useful for BE.
- [x] Refactor Connect auth request routes/views to use the router and router utils (e.g. `routeTrapInside()`) instead of handling certain views as special cases (connect's settings and customs, and transaction details).


src/components/embed/ui/molecules/card/auth-request-card/AuthRequestCard.tsx
```diff
@@ -20,10 +21,24 @@ export function AuthRequestCard({
...
   cancelLabel = browser.i18n.getMessage("cancel"),
   confirmLabel = browser.i18n.getMessage("continue"),
-  isConfirmDisabled,
+  areButtonsDisabled,
   children: childrenProp,
   ...cardProps
 }: AuthRequestCardProps) {
+
+  useEffect(() => {
+    function handleKeyDown(e: KeyboardEvent) {
```
Danziger: This functionality was duplicated in most (but not all) auth request views before.

src/components/embed/ui/organisms/account-selector/AccountSelector.tsx
```diff
@@ -59,7 +57,7 @@ export function AccountSelector({
         content={
           <>
             {wallets.map((wallet, id) => (
-              <DropdownItem key={id} onClick={handleAccountClick}>
```
Danziger, line 62: Bug, but we didn't notice because `setActiveWallet()` had another bug, so it would just throw an error instead of incorrectly unsetting the active wallet.

src/iframe/storage/plasmo-storage/plasmo-storage.mock.ts
```diff
@@ -145,8 +146,9 @@ export class StorageMock extends PlasmoStorage implements StorageMockInterface {
   removeMany: (keys: string[]) => Promise<void> = this.removeItems;
 
   removeAll: () => Promise<void> = () => {
-    return new Promise<void>(() => {
+    return new Promise<void>((resolve) => {
       this.storage.clear();
+      resolve();
```
Danziger, line 151: Bug.

src/routes/embedded/auth-request/connect/connect-custom.view.tsx
```diff
@@ -59,20 +55,10 @@ export function WalletSettingsCustomEmbeddedView() {
     [handlePermissionChange],
   );
 
-  const handleCancel = async () => {
```
Danziger, line 62: Now, accepting or cancelling an auth request will not automatically close the popup. Instead, resolving the last one will redirect users back to `/wallet`, and it will be the SDK who devices if the popup is closed or not (if it was opened automatically, it will; if it was opened before the auth request came in by the user, then it will stay open).

src/routes/embedded/auth-request/connect/connect-settings.view.tsx
```diff
@@ -30,7 +29,7 @@ export function WalletPermissionsRequestEmbeddedView() {
   const { url = "", permissions: authRequestPermissions = [], appInfo = {}, gateway } = authRequest;
 
   const [requestedPermissions, setRequestedPermissions] = useStorage<PermissionType[]>({
-    key: `requested_permissions_${url}`,
```
Danziger, line 33: This stored value is only used to move some data across the 3 different "connect" screens, so there's no need to add `url` in the key because there will never be 2 values at the same time.

src/routes/embedded/auth/restore-shares/keyfile/auth-restore-shares-keyfile.view.tsx
```diff
@@ -43,7 +43,7 @@ export function AuthRestoreSharesKeyfileEmbeddedView({
   });
 
   const areButtonsDisabled = isRecovering || isUploading;
-  const isViewLoading = isUploading;
```
Danziger: Bug (from [the other PR](https://github.com/wanderwallet/Wander/pull/807)).

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1078,6 +1083,52 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
 
     lastUserIdRef.current = userId;
 
+    if (userId && userId !== await PersistentStorage.get<string>(StorageKeys.CONNECT.AUTH.USER_ID)) {
+      // TODO: This is a TEMP FIX to prevent users who share the same device from seeing someone else's tokens and
```
Danziger: @7i7o See here. However, it doesn't seem to work properly, as I only see AO despite calling `loadTokens()` and `updateAoToken()`. Once the page is reloaded, all tokens load fine. Any idea why this might happen?

src/utils/storage/storage.constants.ts
```diff
@@ -0,0 +1,15 @@
+export const SharedStorageKeys = {} as const;
```
Danziger, line 1: Would be good to move all storage keys here at some point.

src/utils/wallets/wallets.service.ts
```diff
@@ -127,6 +127,12 @@ async function fetchFirstAvailableAuthShare(
     for (const wallet of wallets) {
       const { id: walletId, deviceShare } = wallet;
 
+      if (deviceShare === null) {
```
Danziger, line 130: @pawanpaudel93 That issue we had where we would get an error in the console when trying to activate a wallet because its `deviceShare` was `null` happens when the device share has been somehow removed from `localStorage`. With this new check we won't get the error, and this case will be handled by the `return` at the end of this function.

src/wallets/hooks.ts
```diff
@@ -115,21 +115,14 @@ export function useAllWallets() {
 }
 
 export async function setActiveWallet(address?: string) {
-  // verify address
-
-  const [wallets] = useStorage<StoredWallet[]>(
```
Danziger, line 120: Bug: Hook called from a regular function, violates the rules of hooks when called.

src/routes/embedded/auth-request/sign/signDataItem.view.tsx
```diff
@@ -241,7 +208,9 @@ export function EmbeddedSignDataAuthRequestView() {
         </Row>
       )}
 
-      <TransactionMessage transaction={data} />
+      <TransactionMessage
```
pawanpaudel93, line 213: When clicking on Transaction details, it shows a blank screen.

src/utils/embedded/embedded.provider.tsx
pawanpaudel93: I see the tokens were updated to localStorage. When reading the tokens without hooks works but with hooks we don't see the actual tokens. Maybe issue with useStorage hook for Wander Connect usage with localStorage.

src/routes/embedded/auth-request/sign/signDataItem.view.tsx
Danziger: Fixed

7i7o (approved):
tACK 👍

PR #835, opened 2025-05-19
fix: revert nanoid to uuid in keystone

nicholaswma:
Reverted `nanoid` back to `uuid` in Keystone. Keystong QR generator was unable to generate QR codes without the structure provided by `uuid v4`


pawanpaudel93:
@nicholaswma Should this nanoid usage be replaced with uuid as well ? https://github.com/wanderwallet/Wander/pull/835/files#diff-aea8d07c80ece6e885fa5d8801d3b6cc1ff0c3cd4f215ce3f457c93b5689a6b2L130

7i7o (approved):
ACK 👍

PR #837, opened 2025-05-20
Fix dispatch & background error handling

Danziger:
Fix:

1) Dispatch requesting authorization twice if the user rejects the first time.
2) Dispatch retry upload logic not working (apparently, it never did).


src/routes/embedded/auth-request/connect/connect.view.tsx
```diff
@@ -52,7 +52,7 @@ export function EmbeddedConnectAuthRequestView() {
   return (
     <>
       <AuthRequestCard
-        onCancel={rejectRequest}
+        onCancel={() => rejectRequest()}
```
Danziger, line 55: Otherwise, we are passing the click event to the accept/reject handler, which actually have arguments (data or error, respectively).

src/utils/auth/auth.utils.ts
```diff
@@ -270,7 +270,7 @@ export function getPopupResponse<T>(authID: string, popupWindowTabID: number) {
       } else if (isAuthErrorResult(data)) {
         log(LOG_GROUP.AUTH, `auth_result for authID = "${authID}" = Error (${data.error})`);
 
-        reject(data.error);
+        reject(new Error(data.error));
```
Danziger, line 273: Otherwise, background scripts checking for errors might not be able to handle it properly, like it was happening in `dispatch.background.ts`:

```
    if (isError(err) && err.message === ERR_MSG_USER_CANCELLED_AUTH) {
      throw err;
    }
```

src/api/modules/dispatch/allowance.ts
```diff
@@ -23,24 +23,19 @@ export async function ensureAllowanceDispatch(
...
-      );
-    }
+  if (alwaysAsk) {
+    const address = await arweave.wallets.jwkToAddress(keyfile);
 
-    // if (allowance.enabled) {
-    //   await allowanceAuth(appData, allowance, price, alwaysAsk);
-    // }
-  } catch (e) {
-    freeDecryptedWallet(keyfile);
```
Danziger, line 42: `dispatch.background.ts` calls `ensureAllowanceDispatch()` a second time if the first upload attempt fails, so if we call `freeDecryptedWallet()` here, the second one will fail too because the keyfile will no longer be there.

src/api/modules/dispatch/dispatch.background.ts
```diff
@@ -153,6 +157,10 @@ const background: BackgroundModuleFunction<ReturnType> = async (
         type: "BASE",
       },
     };
+  } catch (err) {
+    freeDecryptedWallet(keyfile);
```
Danziger, line 161: With the previous implementation with a single `try-catch`, if the first upload attempt fails, the second one would fail too, because `ensureAllowanceDispatch()` internally had another `try-catch` that would call `freeDecryptedWallet()`, so in the second upload attempt there would be no keyfile.

If we just update `ensureAllowanceDispatch()` to get rid of that internal `try-catch` and call to `freeDecryptedWallet()`, then this single `try-catch` here becomes insecure because if the code inside the `catch` block fails, `freeDecryptedWallet()` might never be called, thus, we need to use 2 separate `try-catch` blocks here.

src/api/modules/dispatch/dispatch.background.ts
```diff
@@ -112,8 +112,21 @@ const background: BackgroundModuleFunction<ReturnType> = async (
...
     if (isError(err) && err.message === ERR_MSG_USER_CANCELLED_AUTH) {
+      freeDecryptedWallet(keyfile);
+
       throw err;
     }
+  }
+
+  try {
+    // grab the user's keyfile again:
+    decryptedWallet = await getActiveKeyfile(appData);
```
Danziger, line 123: With the previous implementation with a single `try-catch`, if the first upload attempt fails, the second one would fail too, because `ensureAllowanceDispatch()` internally has another `try-catch` that calls `freeDecryptedWallet()`, so in the second upload attempt there would be no keyfile. So we need to get the keyfile again in the second attempt.

7i7o (approved):
tACK 👍

PR #839, opened 2025-05-21
feat: add QR code scan to import/recover screens and wallet QR export

pawanpaudel93:
## Summary

This PR adds the QR code scan to wallet import, wallet recover and account recover. It also adds the wallet QR code export to backup.

## How To test
- Make sure to build the sdk and use the built sdk as there is some changes regarding camera allow permission.

Related Jira tickets: [ARC-1268](https://communitylabs.atlassian.net/browse/ARC-1268) [ARC-1288](https://communitylabs.atlassian.net/browse/ARC-1288)

[ARC-1268]: https://communitylabs.atlassian.net/browse/ARC-1268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ARC-1288]: https://communitylabs.atlassian.net/browse/ARC-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wander-connect-sdk/src/components/iframe/iframe.component.ts
```diff
@@ -213,6 +213,7 @@ export class Iframe {
 
     const iframe = document.createElement("iframe");
     iframe.className = "iframe";
+    iframe.allow = "camera *;";
```
pawanpaudel93, line 216: We add camera permissions added to the iframe so new SDK release needs to be made for this.

src/routes/embedded/auth/import-qrcode/components/QrCodeEmbeddedView.tsx
```diff
@@ -0,0 +1,179 @@
...
+  ) : (
+    <OnboardingCard
+      headerText={headerText}
+      subtitle={subtitle}
+      onBackButtonClick={() => navigate(backButtonClickHref)}
+      isLoading={isViewLoading}>
+      {permissionStatus === "granted" ? (
+        <>
+          <QRLoopScanner onResult={handleScanResult} />
+          <Link style={{ color: "#0D6CE9", fontSize: 14, fontWeight: 600, letterSpacing: 0.2 }} href="" target="_blank">
```
pawanpaudel93: Should it be a link or a tooltip showing where to find the QR code?

7i7o (approved):
tACK 👍 

Had some issues (brightness/focus) trying to import from mobile to Connect, it took a while but it worked

src/components/embed/ui/molecules/copy-to-clipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,92 @@
...
+  copySuccess?: string;
+  iconSize?: number;
+  label?: string;
+  labelStyle?: React.CSSProperties;
+  labelAs?: typeof Label;
+  showToast?: boolean;
+  onCopy?: (isCopied: boolean) => void;
+}
+
+const CopyButton = styled.button`
```
Danziger, line 20: For components that are not shared, we should stop using styled components in favor of CSS modules / SCSS, and the theme variables.

src/components/embed/ui/molecules/copy-to-clipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,92 @@
...
+  gap: 8px;
+`;
+
+const Label = styled(Text).attrs({
+  noMargin: true,
+})`
+  overflow: hidden;
+`;
+
+export function CopyToClipboard({
```
Danziger, line 36: There's currently a copy button in the account selector, another one in the backup seedphrase view and another one in the backup QR view. Now that we have this component, they should all use it. I think the root element should be the `InputButton` component though.

src/components/embed/ui/molecules/copy-to-clipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <Label as={labelAs} style={labelStyle}>
+          {label}
+        </Label>
+      )}
+      <Icon as={isCopied ? Check : Copy02} height={iconSize} width={iconSize} $success={isCopied} />
+    </CopyButton>
+  );
+}
+
+const Icon = styled.div<{ height: number; width: number; $success: boolean }>`
```
Danziger, line 88: We should not pass `width`, `height` or `color` to icons. Instead, icons should be `1em` x `1em` and use `currentColor`. This way, the parent will be able to style them as needed.

src/components/embed/ui/molecules/copy-to-clipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export function CopyToClipboard({
+  text,
+  copySuccess,
+  iconSize = 16,
+  label,
+  labelStyle,
+  showToast = true,
+  onCopy,
+  labelAs = Label,
```
Danziger, line 44: Only `text` is needed here. Even after this component is used in all places where we have a copyable feature, all instances should look the same, so there's no need to add plenty of customization options that we are not using, as this can easily lead to unintentional inconsistencies when instantiating this component in different places. Also, this makes this component harder to change/maintain to make sure all props work fine after we do any changes to it, even if those are not used.

src/components/embed/ui/molecules/copy-to-clipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,92 @@
...
+      if (showToast) {
+        toast.error(browser.i18n.getMessage("copy_failed"));
+      }
+    }
+  };
+
+  useEffect(() => {
+    if (!isCopied) return;
+
+    const timeout = setTimeout(() => {
```
Danziger, line 68: Once the props are simplified, this can be called from `copyToClipboard`.

src/components/embed/ui/organisms/qr-loop-scanner/QRLoopScanner.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+        if (cameraPerms.state !== "granted") return;
+      };
+
+      setCameraAllowed(cameraPerms.state === "granted");
+
+      // listen for camera permission changes
+      cameraPerms.addEventListener("change", listener);
+
+      return () => cameraPerms.removeEventListener("change", listener);
```
Danziger: This will not work because the anonymous function is `async`, so it's returning a `Promise`. You need to define the listener outside the IIAF and also return a separate arrow function outside the IIAF with the cleanup code.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }, [decryptedWallet?.keyfile]);
+
+  useEffect(() => {
+    return () => setFrames([]);
+  }, []);
+
+  return (
+    <OnboardingCard
+      headerText={!decryptedWallet ? "Generate QR code" : decryptedWallet?.nickname}
```
Danziger, line 57: I would keep the title always the same and add a label to the input with the address instead. The label should probably just say "Your wallet address" if there's only one, or "<nickname> wallet address" if there are more.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+export function AccountBackupWalletQrCodeEmbeddedView() {
+  const { isDarkMode } = useTheme();
+  const { navigate } = useLocation();
+  const { getDecryptedWallet } = useEmbedded();
+  const [isLoading, setIsLoading] = useState(false);
+  const [copied, setCopied] = useState(false);
+  const [frames, setFrames] = useState<string[]>([]);
+  const [decryptedWallet, setDecryptedWallet] = useState<LocalWallet<JWKInterface> | null>(null);
+
+  const fetchDecryptedWallet = useCallback(async () => {
```
Danziger, line 27: This function and the 2 `useEffects` below can be simplified in a single `useEffect`:

```
useEffect(() => {
  let decryptedWallet: JWKInterface | undefined;

  async function decryptWallet() {
    try {
      setIsLoading(true);
      decryptedWallet= await getDecryptedWallet();
      setFrames(dataToFrames(JSON.stringify(decryptedWallet.keyfile)));
    } catch (error) {
      toast.error("Error getting wallet");
    } finally {
      setIsLoading(false);
      if (decryptedWallet) freeDecryptedWallet(decryptedWallet.keyfile);
    }
  }

  decryptWallet();

  return () => {
    if (decryptedWallet) freeDecryptedWallet(decryptedWallet.keyfile);
  };
}, [getDecryptedWallet]);
```

If this `useEffect` is being invoked multiple times, you can add a `ref` to make sure it only runs once.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { QRCodeLoop } from "~components/QRCodeLoop";
+import { useTheme } from "~components/embed/contexts/ThemeContext";
+
+export function AccountBackupWalletQrCodeEmbeddedView() {
+  const { isDarkMode } = useTheme();
+  const { navigate } = useLocation();
+  const { getDecryptedWallet } = useEmbedded();
+  const [isLoading, setIsLoading] = useState(false);
+  const [copied, setCopied] = useState(false);
+  const [frames, setFrames] = useState<string[]>([]);
```
Danziger, line 24: I think it's better to use an object with number keys in here (`Record<number, string>`), as we should use a method similar to `freeDecryptedWallet()` for it.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  useEffect(() => {
+    if (decryptedWallet?.keyfile) {
+      setFrames(dataToFrames(JSON.stringify(decryptedWallet.keyfile)));
+      freeDecryptedWallet(decryptedWallet.keyfile);
+    }
+  }, [decryptedWallet?.keyfile]);
+
+  useEffect(() => {
+    return () => setFrames([]);
```
Danziger, line 52: This doesn't remove the old values from memory, as you are just creating a new array/reference, and the other one will still be around until the GC takes care of it. I guess the idea here was to implement something similar to `freeDecryptedWallet()`, so I think using an object would be better. That way, what we would be doing is keeping the same object/reference but mutating its props/values.

Still, I have many doubts about this given 1) that the same issue with new references being created apply when changing the props/values and 2) we are now hooked into React, so updating the object that's also inside the state might cause some issues or not work as expected, but it was part of the last audit, so let's give it a try anyways.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }, []);
+
+  return (
+    <OnboardingCard
+      headerText={!decryptedWallet ? "Generate QR code" : decryptedWallet?.nickname}
+      onBackButtonClick={() => navigate("/account/backup-wallet")}
+      isLoading={isLoading}>
+      <Flex direction="column" gap={32} justify="center" align="center">
```
Danziger, line 60: I think there's no need for this `Flex` here, as `OnboardingCard` takes care of evenly spacing its children. I would postpone fine adjustments until we have a more stable version of the UI. Then we can see if it makes sense to space out further certain elements, but that should be handled using the theme spacing variables and in a reproducible way if possible. Also, adding some CSS into the views should be preferred to using `Flex` / `Box` components that create unnecessary HTML elements (my intention is actually to delete those).

src/components/QRCodeWrapper.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from "styled-components";
+import { IS_EMBEDDED_APP } from "~utils/embedded/embedded.constants";
+
+export const QRCodeWrapper = styled.div<{ size?: number }>`
```
Danziger, line 4: Unless `QRCodeLoop` is used somewhere without `QRCodeWrapper`, I think `QRCodeWrapper` should just be part of `QRCodeLoop`.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+              backgroundColor: "#fff",
+              padding: "12px",
+              borderRadius: "12px",
+              height: 299,
+              width: 299,
+            }}>
+            {frames.length > 0 && <QRCodeLoop frames={frames} fps={5} size={275} />}
+          </div>
+        </QRCodeWrapper>
+        <Flex
```
Danziger, line 73: We also have multiple instances of an input field with a copiable (non-secret) value (e.g. in Export wallet and Download recovery file views). I suggest you create a new `CopiableAddressInput` and reimplement by combining `TextInput` (readonly) + `CopyToClipboardButton`.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+          direction="column"
+          justify="center"
+          align="center"
+          gap={12}
+          padding={12}
+          style={{ background: isDarkMode ? "#2C2C2E" : "#EBEBF0", borderRadius: 8, width: "100%" }}>
```
Danziger, line 79: Avoid using JS for theme-related changes (`isDarkMode`). Also, all these values should come from the theme. In any case, this comment is irrelevant if you implement the change in the comment above.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+              fontSize: 14,
+              fontWeight: 500,
+            }}>
+            {decryptedWallet?.address}
+          </span>
+          <CopyToClipboard
+            onCopy={setCopied}
+            showToast={false}
+            label={browser.i18n.getMessage(copied ? "copied" : "copy")}
+            labelAs={({ children }) => <Text variant="bodySm">{children}</Text>}
```
Danziger, line 95: As mentioned in the `CopyToClipboard` comments, this will look much simpler and become easier to use with way less props. Also, there's no reason for the parent to manage the `copied` / `setCopied` state.

src/routes/embedded/account/backup-wallet/backup-wallet-qrcode.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [copied, setCopied] = useState(false);
+  const [frames, setFrames] = useState<string[]>([]);
+  const [decryptedWallet, setDecryptedWallet] = useState<LocalWallet<JWKInterface> | null>(null);
+
+  const fetchDecryptedWallet = useCallback(async () => {
+    try {
+      if (decryptedWallet) return;
+      setIsLoading(true);
+      const wallet = await getDecryptedWallet();
```
Danziger, line 31: Using a seedphrase instead of a keyfile (when available) would mean this works with a single QR, but not sure if that's good or bad (easier to peak over someone's shoulder and steal it). 

Maybe just adding the seedphrase inside the keyfiles would be sufficient in order to allow users to preserve access to them.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -304,6 +312,28 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
     [walletId, walletAddress],
   );
 
+  const getDecryptedWallet = useCallback(async (): Promise<LocalWallet<JWKInterface>> => {
+    log(LOG_GROUP.EMBEDDED_FLOWS, `getDecryptedWallet()`);
+
+    const decryptedWallet = (await getKeyfile(walletAddress)) as LocalWallet<JWKInterface>;
+
+    try {
+      const { wallet: updatedWallet } = await WalletService.registerWalletExport({
```
Danziger, line 346: This is probably a bad idea, because seeing the QR code doesn't guarantee the user has a copy of it. I think we need to update the backend to include a new `QR` type for this, and either make that type not update the export count (so that we keep prompting users to back up) or include a download button that downloads the QR as GIF/video. In that case, we only register this when they click download, not when we decrypt the keyfile. That means users should be able to upload that QR code too. In that case, it might be easier to parse if we include the JWK JSON in it as metadata (wondering if any upload service like Drive or Dropbox would automatically get rid of that metadata).

7i7o (approved):
tACK 👍

PR #840, opened 2025-05-21
feat: add pi as default token & remove qAR default token

pawanpaudel93:
## Summary

This PR adds the PI as default token to new & existing users and also it fixes PI token to top like AR and AO tokens. It also removes the `qAR` token from default tokens.


pawanpaudel93:
@7i7o Should we fix `PI` token at the top along with `AR` & `AO` so that in the dashboard PI token is not sorted based on the token balance or fiat balance ?

7i7o:
I don't think so. Let's discuss with Matt, but IMO there might be people who don't opt in for PI and might not enjoy having it at they top as default

src/tokens/token.ts
```diff
@@ -63,17 +63,21 @@ export async function loadTokens() {
...
+    defaultTokens[0], // AR
+  ];
 
-  if (!aoTokens.some((t) => t.processId === ar.processId)) {
-    aoTokens.unshift(ar);
+  for (let i = 0; i < requiredTokens.length; i++) {
+    const token = requiredTokens[i];
+    if (!existingProcessIds.has(token.processId)) {
+      aoTokens.unshift(token);
+    }
```
Danziger, line 80:
```suggestion
  requiredTokens.forEach((requiredToken) => {
    if (!existingProcessIds.has(requiredToken.processId)) {
      aoTokens.unshift(requiredToken);
    }
```

src/tokens/token.ts
```diff
@@ -63,17 +63,21 @@ export async function loadTokens() {
...
   }
```
Danziger:
```suggestion
  requiredTokens.forEach((requiredToken) => {
    if (!existingProcessIds.has(requiredToken.processId)) {
      aoTokens.unshift(requiredToken);
    }
  });
```

Danziger (approved):
👍 utACK

PR #844, opened 2025-05-22
Fix issue displaying long token names and/or balances in the tokens list

Danziger:
The `AssetItem` code has some small issues:

- Magic numbers.
- Some duplicate code.
- Size calculations are not 100% accurate, they seem to be a few pixels off.

But we can improve it later.

Right now it already ajusts its layout dynamically to accommodate long token names and long balances as best as possible. For example, "Permaweb Index Token" wouldn't fit before, even with no balance, but now it does:

For slightly longer content, we try to split the available space 50/50 between the token name and the token balance, while keeping the ticker intact:

For even longer content, both token names and token balances collapse, but hovering over them reveals them:

For even longer content that cannot fit the screen width, a marquee event is applied on hover:


7i7o (approved):
tACK 👍

PR #850, opened 2025-05-23
refactor: update aoprocess and use new cu

pawanpaudel93:
## Summary

This PR updates the following:

- Update `AOProcess` class based on [ao-process.ts](https://github.com/ar-io/ar-io-sdk/blob/main/src/common/contracts/ao-process.ts)
- Use `https://cu.ardrive.io` CU based on [CHANGELOG.md](https://github.com/ar-io/ar-io-sdk/blob/09dabee5fff189cf4419e9661bf04a2908d453a3/CHANGELOG.md?plain=1#L28)


7i7o (approved):
tACK 👍

Danziger:
@pawanpaudel93 @7i7o Why are we manually copying their class instead of installing the official package?

pawanpaudel93:
@Danziger We weren't using all the functionality provided by the SDK and when I tried to use the SDK, the build size of BE increased significantly.

Danziger (approved):
Let's merge it as it is but we should look into the issue with the bundle size, as it would be better to have this listed as a dependency in `package.json` rather than manually copying the file(s) we need.

PR #856, opened 2025-05-29
fix: ARC-1363: Auth event missing with email and password

Danziger:
This PR includes:

- Move logic to complete OAuth authentication in the popup window to `main.ts` and avoid mounting the React app in that case, as we only need to validate the session, post a message to the iframe, and close the popup window. This should make the whole flow slightly faster.
- Add a spinner directly into the `index.html` file.
- Fix iframe styles (in the SDK) and `popup.css` styles so that the iframe's content covers all the available space. This way, the `#cover` element, which now contains the spinner, always covers everything. Also, some logic has been included so that the the background color matches that of the OAuth provider, so that the redirect is less noticeable to users.
- Remove `AuthErrorEmbeddedView`. Errors are now handled directly in the view that initiated the flow that cause them (e.g. Auth view).
- Extract friendly error messages from auth errors.
- Close https://communitylabs.atlassian.net/browse/ARC-1363.
- Close https://communitylabs.atlassian.net/browse/ARC-1313.
- Close https://communitylabs.atlassian.net/browse/ARC-1292.


src/iframe/iframe.html
Danziger, line 1: Duplicated file (unused).

src/iframe/main.tsx
```diff
@@ -7,8 +7,23 @@ import {
...
+  BACKGROUND_COLORS_BY_PROVIDER,
+  OAUTH_ERROR_MSG_TYPE,
+  OAUTH_SUCCESS_MSG_TYPE,
+  OAuthError,
+  POPUP_CONFIRMATION_TIMEOUT_MS,
+  POPUP_ON_AUTH_TIMEOUT_MS,
+} from "~utils/authentication/authentication.utils";
+import { jwtDecode } from "jwt-decode";
 
-import "../../assets/popup.css";
```
Danziger, line 11: Already included from `index.html`.

src/utils/authentication/authentication.service.ts
```diff
@@ -1,39 +1,192 @@
...
     },
   });
 
   if (error) throw error;
 
-  return { url: data.url };
+  const { url } = data;
+
+  if (!url) {
+    throw new Error(`Missing OAuth URL.`);
```
Danziger, line 47: TODO: Extract to constants.

src/iframe/main.tsx
```diff
@@ -21,4 +36,100 @@ if (process.env.NODE_ENV === "development") {
...
+
+      // If the iframe doesn't close this popup window after receiving the
+      // AUTH_COMPLETE_MSG_TYPE message, it will close itself after POPUP_CONFIRMATION_TIMEOUT_MS:
+
+      setTimeout(() => window.close(), POPUP_CONFIRMATION_TIMEOUT_MS);
+    });
+  }
+
+  completeOAuthAuthentication();
+} else if (window.location.search.startsWith("?error=")) {
```
Danziger, line 107: Instead of showing the error in the popup window, we send it to the iframe and close the popup window immediately. The error is then shown in the iframe.

src/routes/embedded/auth/auth-email-signup/auth-email-signin.view.tsx
```diff
@@ -29,77 +29,38 @@ export function AuthEmailSigninEmbeddedView() {
...
-          toast.error(error.message);
-          if (error.code === "email_not_confirmed") {
-            navigate(EmbeddedPaths.AuthEmailVerify);
-          }
-          return;
-        }
-
-        const interval = setInterval(() => {
-          if (authStatusRef.current === "unlocked") {
-            navigate(EmbeddedPaths.WalletHomeEmbeddedView);
```
Danziger, line 76: No redirect to home. The router takes care of this once `authStatus` is updated.

src/routes/embedded/auth/auth-email-signup/auth-email-verify.view.tsx
```diff
@@ -1,33 +1,38 @@
...
 export function AuthEmailVerifyEmbeddedView() {
   const { navigate } = useLocation();
-  const { authStatus } = useEmbedded();
+  const { authenticate } = useEmbedded();
   const { email } = useSearchParams<{ email: string }>();
-  const [isResending, setIsResending] = useState(false);
-  const [isVerifying, setIsVerifying] = useState(false);
   const [cooldownTime, setCooldownTime] = useState(0);
   const [canResend, setCanResend] = useState(false);
-  const [otpDigits, setOtpDigits] = useState<string[]>(Array(OTP_LENGTH).fill(""));
```
Danziger, line 25: Now using refs instead of state.

src/routes/embedded/auth/auth-email-signup/auth-email-verify.view.tsx
```diff
@@ -61,159 +66,150 @@ export function AuthEmailVerifyEmbeddedView() {
...
-      return;
-    }
+    if (!input) return;
+
+    input.focus();
 
-    handleVerifyCode(otpCode);
-  }, [email, otpDigits]);
+    requestAnimationFrame(() => {
+      input.select();
```
Danziger, line 115: To avoid weird/inconsistent behavior depending on cursor position, we just select the input value/content.

src/routes/embedded/auth/auth-email-signup/auth-email-verify.view.tsx
```diff
@@ -61,159 +66,150 @@ export function AuthEmailVerifyEmbeddedView() {
...
+    });
+  }, []);
 
-  const handlePaste = useCallback(
-    (e: React.ClipboardEvent<HTMLInputElement>, index: number) => {
+  const handleInputOrPaste = useCallback(
```
Danziger, line 119: Single handler for `onInput` and `onPaste`.

src/routes/embedded/auth/auth-email-signup/auth-email-verify.view.tsx
```diff
@@ -61,159 +66,150 @@ export function AuthEmailVerifyEmbeddedView() {
...
 
-        // Find next empty input or focus last input
-        const nextEmptyIndex = newDigits.findIndex((digit, idx) => idx >= index && !digit);
-        const focusIndex = nextEmptyIndex !== -1 ? nextEmptyIndex : OTP_LENGTH - 1;
-        focusInput(focusIndex);
+      let nextEmptyIndex = currentIndex;
 
-        setOtpDigits(newDigits);
-        autoSubmitIfComplete(newDigits);
+      if (sanitizedValue.length === OTP_LENGTH) {
```
Danziger, line 133: If pasting a whole code, take it all, ignoring in what `input` the paste took place.

src/routes/embedded/auth/auth-email-signup/auth-email-verify.view.tsx
```diff
@@ -61,159 +66,150 @@ export function AuthEmailVerifyEmbeddedView() {
...
+
+      handleVerifyCode();
     },
-    [otpDigits, focusInput, autoSubmitIfComplete],
+    [focusInput, handleVerifyCode],
   );
 
   const handleKeyDown = useCallback(
-    (e: React.KeyboardEvent<HTMLInputElement>, index: number) => {
+    (e: React.KeyboardEvent<HTMLInputElement>) => {
```
Danziger, line 162: Added support for up, down and delete keys here.

src/utils/authentication/authentication.service.ts
```diff
@@ -1,39 +1,190 @@
...
+  // Calculate center position for the popup:
+
+  const width = Math.min(400, document.documentElement.offsetWidth);
+  const height = Math.min(600, window.screen.availHeight - 32);
+  const left = window.screenX + (window.outerWidth - width) / 2;
+  const top = window.screenY + (window.outerHeight - height) / 2;
+
+  // Open the provider's OAuth page (Google, Twitter, Facebook, Apple) in a popup window:
+
+  let popup = window.open(
```
Danziger, line 59: Better to handle the whole flow in the service. `AuthenticationService.authenticate()` now returns a `Promise` that resolves once the auth popup window has been opened, the authentication there has happend, and the `OAUTH_SUCCESS_MSG_TYPE` (or error) message is received by the iframe.

wander-connect-sdk/src/components/iframe/iframe.template.ts
```diff
@@ -63,8 +63,8 @@ export const getIframeTemplateContent = ({ customStyles, cssVariableKeys = [] }:
       box-shadow: var(--boxShadow);
       width: calc(var(--preferredWidth) + 2 * var(--borderWidth));
       height: calc(var(--preferredHeight) + 2 * var(--borderWidth));
-      min-width: 400px;
-      min-height: 400px;
+      min-width: calc(400px + 2 * var(--borderWidth));
```
Danziger, line 66: So that the minimum size of 400x400px is applied to the iframe content.

assets/popup.css
```diff
@@ -72,19 +72,34 @@
...
+}
+
+html[data-theme="light"] {
+  --defaultBackgroundColor: white;
+  --defaultTextColor: black;
+}
+
+html[data-theme="dark"] {
+  --defaultBackgroundColor: black;
+  --defaultTextColor: white;
```
7i7o: This change is great, but it's breaking the theme in BE because `data-theme` is not set for all BE screens.
We can modify `~/assets/popup.js` so this .css works, retrieving the theme from the extension's local storage with something like this:

```js
async function setTheme() {
  const result = await chrome.storage.local.get(['setting_display_theme']);
  let theme = result.setting_display_theme || 'system';
  if (typeof theme === 'string' && theme.startsWith('"')) 
    theme = JSON.parse(theme);
  document.documentElement.dataset.theme = theme;
}

setTheme();
```

The only downside with this approach is that BE doesn't receive the theme as a param, so the async nature of retrieving from extension storage makes this slower than the iframe param-based approach.

7i7o (changes requested):
WC is working great 👍 

We just need to make a small change in `~/assets/popup.js` in order for themes to be fixed for BE

Look at my comment in the PR

With a fix for the BE theme, we can merge

assets/popup.css
Danziger: Fixed, but no need to change the JS. I messed up the selectors so BE was not getting the right CSS variables applied.

PR #857, opened 2025-05-29
fix: ARC-1355: Safari issue with `pointer-events: none`

Danziger:
Removed `pointer-events: none` and animated both `opacity` and `display` using just transitions with `transition-behavior` and `@starting-style`. These are relatively new but we should be good in terms of browser support. See:

- https://caniuse.com/mdn-css_properties_transition-behavior
- https://caniuse.com/mdn-css_at-rules_starting-style

Closes https://communitylabs.atlassian.net/browse/ARC-1355


nicholaswma (approved):
tACK! thank you!

PR #860, opened 2025-06-05
fix: ARC-1372: Fix auth popup closing

Danziger:


Danziger:
This needs some work to be properly fixed. We'll have to check if it's possible to use long-lived connections with `webext-bridge` or otherwise implement some kind of handshake between popup and background to avoid the race condition between the popup `window.close()` call and the logic in `createAuthPopup()` to open a new popup or reuse the currently open (that might already be in the process of being closed) one.

PR #862, opened 2025-06-11
feat: ao yield agents

pawanpaudel93:


src/components/SliderMenu.tsx
```diff
@@ -68,7 +68,7 @@ export default function SliderMenu({
   return createPortal(<AnimatePresence>{contentElement}</AnimatePresence>, document.body);
 }
 
-const ExitButton = styled(CloseIcon)`
+export const ExitButton = styled(CloseIcon)`
```
Danziger: This should be extracted as its own component.

src/components/popup/Navigation.tsx
```diff
@@ -86,10 +89,17 @@ export const NavigationBar = () => {
             active={active}
             data-active={active ? "true" : "false"}
             key={index}
-            onClick={() => navigate(button.route as WanderRoutePath)}>
+            onClick={() => {
+              if (button.route === "/agents") {
+                trackEvent(EventType.AGENT_DASHBOARD, {});
```
Danziger, line 94: Probably not relevant for this PR, as I see this is already inconsistent across the whole repo, but I see some `trackEvent` being called before `navigate` while others are called from `useEffect()`s. I prefer the latter because it's a more reliable way to make sure a specific event is triggered every time a specific view is render, regardless or how the user got there. Even better than manually adding `useEffect()`s across some views, we could move this to the router config object and `Routes` / `Page` components.

src/routes/popup/agents/ao-yield/agent-activated.tsx
```diff
@@ -0,0 +1,89 @@
...
+import agentActivated from "url:/assets/agents/images/agent-activated.svg";
+import agentNotActivated from "url:/assets/agents/images/agent-not-activated.svg";
+import { useLocation, useSearchParams } from "~wallets/router/router.utils";
+import { PopupPaths } from "~wallets/router/popup/popup.routes";
+import { useEffect } from "react";
+import { trackPage, PageType } from "~utils/analytics";
+
+export function AOYieldAgentActivatedView() {
+  const { navigate } = useLocation();
+  const { activationStatus = "success" } = useSearchParams<{ activationStatus: "success" | "error" }>();
```
Danziger, line 14: It looks like the 2 different variants of this page only share the structure of the HTML/JSX. It would be better to create 2 separate views with 2 different routes (instead of using search params) and either duplicate the JSX or extract it to a shared layout component if it's used in more places (e.g. similar to the `OnboardingCard` or `Card` components).

src/routes/popup/agents/ao-yield/confirm-agent.tsx
```diff
@@ -0,0 +1,250 @@
...
+      <Wrapper>
+        <Content>
+          <Flex
+            align="center"
+            justify="space-around"
+            style={{
+              maskImage: "linear-gradient(to bottom, transparent, black 20%, black 80%, transparent)",
+              WebkitMaskImage: "linear-gradient(to bottom, transparent, black 20%, black 80%, transparent)",
+            }}>
+            <SvgImageWithBackground
```
Danziger, line 157: This pattern seems to repeat in different components (`SvgImageWithBackground` + 1 or 2 `img`). Maybe we should create a component for it.

src/routes/popup/agents/ao-yield/confirm-agent.tsx
```diff
@@ -0,0 +1,250 @@
...
+                  {i === 1 && aoYieldAgent?.asset?.logo && <img src={aoYieldAgent.asset.logo} height={18} width={18} />}
+                </PropertyValue>
+              </TransactionProperty>
+            ))}
+          </Flex>
+          {askPassword && transferRequirePassword && (
+            <Input
+              placeholder="Password"
+              sizeVariant="small"
+              {...passwordInput.bindings}
```
Danziger, line 196: Use refs when possible. Here, the value of the input is not used until `handleActiveAgent` is called, so using a ref to read the value directly from the input is a superior option (causes no renders).

src/routes/popup/agents/ao-yield/confirm-agent.tsx
```diff
@@ -0,0 +1,250 @@
...
+              label={browser.i18n.getMessage("enter_password_confirm")}
+              labelStyle={{ marginBottom: -4, color: theme.primaryText, fontSize: 16, fontWeight: 500 }}
+              type="password"
+              onKeyDown={async (e) => {
+                if (e.key !== "Enter") return;
```
Danziger: We should not do that. The whole or most of the page should be wrapped in a `<form>`, and `handleActiveAgent` should be called `onSubmit`. Forms handle pressing `Enter` automatically, so no need for `onKeyDown`.

Also, the `Button` below should use `type="submit"`. With that, no `onClick` needed either. Clicking the button will submit the form too.

For reference, see `src/routes/embedded/auth/auth/auth.view.tsx`. If the components used for BE do not support this and too many changes are needed, you can leave it as it is, but please add an issue to the backlog to address it later.

src/routes/popup/agents/ao-yield/confirm-agent.tsx
```diff
@@ -0,0 +1,250 @@
...
+        runningTime,
+        dailyConversionPercentage: aoYieldAgent.conversionPercentage,
+      });
+
+      // Schedule swap execution
+      scheduleSwapExecution();
+
+      navigate(PopupPaths.AOYieldAgentActivated, { search: { activationStatus: "success" } });
+
+      TempTransactionStorage.remove("ao-yield-agent");
```
Danziger, line 132: If this fails, the data is not removed. Is this intentional?

src/routes/popup/agents/ao-yield/create-agent.tsx
```diff
@@ -68,9 +80,33 @@ export function CreateAOYieldAgentView() {
...
+      asset: selectedAsset,
+      slippage: selectedSlippage,
+      runIndefinitely,
+      startDate: startDate.getTime(),
+      endDate: endDate.getTime(),
+    });
+    navigate(PopupPaths.ConfirmAOYieldAgent);
+  };
+
+  useEffect(() => {
```
Danziger: Derived state, so this should probably use `useMemo` rather than `useState` + `useEffect`.

src/routes/popup/agents/ao-yield/edit-agent.tsx
```diff
@@ -0,0 +1,302 @@
...
+        content: browser.i18n.getMessage("error_updating_agent"),
+        type: "error",
+        duration: 2400,
+      });
+    } finally {
+      setIsLoading(false);
+    }
+  }
+
+  useEffect(() => {
```
Danziger: Same as above. This should use `useMemo()`.

src/routes/popup/agents/ao-yield/edit-agent.tsx
```diff
@@ -0,0 +1,302 @@
...
+  const closeDateSelector = () => {
+    setShowDateSelector(false);
+  };
+
+  const handleDateSelect = (selectedStartDate: Date, selectedEndDate: Date) => {
+    setStartDate(selectedStartDate);
+    setEndDate(selectedEndDate);
+  };
+
+  const formatDateRange = (type: "start" | "end" = "start") => {
```
Danziger: Having explicit params help others easily understand the code. I would define this function outside the component and pass the date we want to format and the fallback label as params. Something like this:

```
function formatDate(date: Date | null, fallbackLabel: string) {
  return date?.toLocaleDateString("en-US", {
      month: "short",
      day: "numeric",
      year: "numeric",
  }) || fallbackLabel;  
}
```

Also, why are we using `dayjs` in some places but not in others?

src/routes/popup/agents/components/AOMintingStatusModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+import type { MintingStatus } from "~utils/agents/types";
+
+interface AOMintingStatusModalProps {
+  mintingStatus: MintingStatus;
+  isError: boolean;
+}
+
+export function AOMintingStatusModal({ mintingStatus, isError }: AOMintingStatusModalProps) {
+  const [open, setOpen] = useState(true);
+  const status = useMemo(() => (mintingStatus === "Paused" ? "Paused" : "Resumed"), [mintingStatus]);
```
Danziger: No need to use memo for inexpensive logic.

src/routes/popup/agents/components/AOYieldAgentListItem.tsx
```diff
@@ -47,11 +70,22 @@ export const AOYieldAgentListItem = ({ aoYieldAgentAvailable }: { aoYieldAgentAv
...
+          />
         </Flex>
       }
       active
-      onClick={() => navigate(PopupPaths.CreateAOYieldAgent)}
+      onClick={() =>
+        aoAgent.status === "Active"
+          ? navigate(PopupPaths.ManageAOYieldAgent)
+          : isHistory
+            ? navigate(PopupPaths.AOYieldAgentInfo.replace(":id", aoAgent.id) as WanderRoutePath)
```
Danziger: Maybe better to update `navigate()` and `NavigateOptions` so that we can pass the params:

```
  navigate(PopupPaths.AOYieldAgentInfo, { params: { id: aoAgent.id } })
```

src/routes/popup/agents/components/ao-yield/DateSelectorModal.tsx
```diff
@@ -11,16 +11,47 @@ interface DateSelectorModalProps {
   startDate: Date | null;
   endDate: Date | null;
   onSelect: (startDate: Date, endDate: Date) => void;
+  runIndefinitely: boolean;
 }
 
-export function DateSelectorModal({ open, onClose, onSelect, startDate, endDate }: DateSelectorModalProps) {
+const monthNames = [
```
Danziger, line 17: These won't be translated.

src/routes/popup/agents/components/ao-yield/DateSelectorModal.tsx
```diff
@@ -30,9 +61,16 @@ interface DateSelectorScreenProps {
   onSelect: (startDate: Date, endDate: Date) => void;
   currentStartDate: Date | null;
   currentEndDate: Date | null;
+  runIndefinitely: boolean;
 }
 
-const DateSelectorScreen = ({ onClose, onSelect, currentStartDate, currentEndDate }: DateSelectorScreenProps) => {
+const DateSelectorScreen = ({
```
Danziger, line 75: Not sure implementing a custom date selector is a good idea. Probably easier and safer to find a library we can customize to fit our needs.

src/routes/popup/agents/components/ao-yield/agent-info.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const { data: agentInfo } = useAOYieldAgentInfo(agentId);
+  const { navigate } = useLocation();
+
+  const asset = useMemo(() => {
+    if (!agent) return null;
+    return assets.find((asset) => asset.id === agent.tokenOut);
+  }, [agent]);
+
+  const agentDetails = useMemo(() => {
+    if (!agent || !asset) return [];
```
Danziger: Very similar logic inside `src/routes/popup/agents/ao-yield/confirm-agent.tsx`. Maybe this could be extracted to a util function or hook?

PR #863, opened 2025-06-12
feat: agents

pawanpaudel93:


src/components/Slider.tsx
```diff
@@ -0,0 +1,161 @@
...
+      <SliderWrapper>
+        {showLabels && (
+          <SliderLabel position="left">
+            <Text variant="secondary" size="sm" weight="medium" noMargin>
+              {minLabel || min}
+            </Text>
+          </SliderLabel>
+        )}
+        <SliderTrackContainer>
+          <SliderInput
```
Danziger: Using a controlled input for sliders make them sluggish/slow. A better implementation would:

- Use an uncontrolled range input for a crispier response, with an `inputRef` prop so that the parent can read the value when needed.
- Move `value` to a state inside the component, as there's no need to pass values up and down if the parent component doesn't need to do anything with it. Updates to this state should be wrapped in a React transition. Alternatively, just use refs for the other elements (`SliderFill`, `SliderThumb`, label... and update those manually using `requestAnimationFrame()`).
- Consider also throttle calls to `onChange` so that updates to the parent have a cap on how fast/often they propagate.

src/routes/popup/agents/components/LiquidOpsAgentListItem.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { useMemo, type ComponentProps } from "react";
+import BigNumber from "bignumber.js";
+import { formatNumber } from "../liquidops/utils/format";
+import { EventType, trackEvent } from "~utils/analytics";
+
+interface Props {
+  activeTokens: ActiveAgentToken[];
+}
+
+export const LiquidOpsAgentListItem = ({ activeTokens }: Props) => {
```
Danziger, line 21: Hover styles for the whole Liquid Ops Agent item is confusing. When hovering, the whole item background changes, but the user cannot differentiate when the click is going to go to the  "Show more" or to the children items. See:

image

assets/ecosystem/war.svg
Danziger, line 1: These SVGs just contain a regular image, so it would be better to just use some other format (png, jpg, webp...).

src/routes/popup/agents/components/ao-yield/SlippageSelectorModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setSelectedSlippage(Math.min(newValue, 10));
+  }
+
+  return (
+    <Flex direction="column" gap={24} height="100%" width="100%">
+      <Text variant="secondary" size="sm" weight="medium" noMargin>
+        {browser.i18n.getMessage("slippage_description")}
+      </Text>
+      <Flex direction="row" gap={16} align="center" justify="center" padding="20px 0" width="100%">
+        <RoundedButton onClick={handleDecreaseSlippage}>-</RoundedButton>
```
Danziger: Clicking this multiple times could display many toasts with the same message:

image

Danziger: Also, it would be good if:

- Holding the buttons down was possible. See the ones here: https://andorratechvalley.com/
- The text element had a fixed with so that the - and + buttons do not move while changing the value.
- The characters were properly centered inside the circles.

src/routes/popup/agents/ao-yield/create-agent.tsx
```diff
@@ -0,0 +1,292 @@
...
+                  -
+                </Text>
+                <InputButton
+                  style={{ background: theme.surfaceTertiary }}
+                  body={
+                    <Text size="lg" weight="medium" noMargin>
+                      {runIndefinitely ? "∞" : formatDate(endDate, "End date")}
+                    </Text>
+                  }
+                  onClick={openDateSelector}
```
Danziger: When clicking the end date button, the date selector should open with the end date input selected.

Also, if run indefinitely is checked, the end date button should be disabled and not open the selector when clicked.

src/routes/popup/agents/components/ao-yield/DateSelectorModal.tsx
```diff
@@ -0,0 +1,792 @@
...
+  );
+
+  const isDateInRange = (day: number) => {
+    if (!startDate || !endDate) return false;
+    const date = new Date(currentDate.getFullYear(), currentDate.getMonth(), day);
+    // Include both start and end dates, even if they're the same
+    return date >= startDate && date <= endDate;
+  };
+
+  const isDateSelected = (day: number) => {
```
Danziger: Most of these functions could be defined in a separate utils file by simply adding some extra args to them, like:

```
  export interface DateCheckOptions {
    day: number;
    currentDate: Date;
    startDate: Date;
    endDate: Date;
  }

  export function isDateSelected({
    day,
    currentDate,
    startDate,
    endDate,
  }) {
    const date = new Date(currentDate.getFullYear(), currentDate.getMonth(), day);
    return (startDate && date.getTime() === startDate.getTime()) || (endDate && date.getTime() === endDate.getTime());
  }
```

When calling these functions, you only need to create one single `DateCheckOptions` object, so the calls still have one simple param.

This will make this component easier to read, facilitate writing tests for these functions (eventually) and make the component faster to render, as these functions are not re-defined on each render.

src/routes/popup/agents/components/ao-yield/DateSelectorModal.tsx
```diff
@@ -0,0 +1,792 @@
...
+            onClick={() => navigateMonth(-1)}
+            $disabled={
+              currentDate.getFullYear() <= new Date().getFullYear() && currentDate.getMonth() <= new Date().getMonth()
+            }>
+            <ChevronLeft width={20} height={20} color={theme.primaryText} />
+          </NavButton>
+          <Text size="lg" weight="semibold" noMargin>
+            {monthNames[currentDate.getMonth()]} {currentDate.getFullYear()}
+          </Text>
+          <NavButton onClick={() => navigateMonth(1)}>
```
Danziger, line 501: Similarly to my comment in the range input, it would be good to be able to hold down these buttons to quickly move back/forward, and/or to be able to change the month/year manually (assuming going further into the future is a common scenario).

src/routes/popup/agents/components/ao-yield/DateSelectorModal.tsx
Danziger, line 1: Some general feedback of possible improvements for later:

- There's some dead space in-between days where clicks won't do anything. It would be good to eliminate that by using before/after CSS pseudo-elements to expand the interactable area of the day buttons.

- It would be nice to display the range selection when hovering over the days instead of waiting for the click, as it's easier for users to visualize the new updated range that way.

- When clicking the start or end dates in the selector, it would be good to move to that month.

- There seems to be a dark grey circle leaking into the purple end cap here:

    image

- If you click the start date button outside the selector, change the start date and then close it, the dates won't update. Maybe it would be good to add a "Save" button instad, as usually I would expect the close button to cancel the changes and have a different button to persist them.

- It might be better to remove the run indefinitely check from outside and add 2 special values inside the date picker. "Today" to easily set the start date to today, and "Indefinitely"/"Forever" for the end date.

src/routes/popup/agents/ao-yield/create-agent.tsx
```diff
@@ -0,0 +1,292 @@
...
+              }
+              outerLabel
+            />
+          </Flex>
+        </Content>
+        <Flex gap={8}>
+          <Button style={{ flex: 1 }} disabled={continueButtonDisabled} onClick={handleContinue} fullWidth>
+            {browser.i18n.getMessage("continue")}
+          </Button>
+          <Button
```
Danziger, line 234: If you fill in the other fields and then click this button (as it's hard to know what it means), when coming back, all the fields are empty. Adding a tooltip to the button could help.

src/routes/popup/agents/ao-yield/confirm-agent.tsx
```diff
@@ -0,0 +1,234 @@
...
+            name: "Token-Out",
+            value: aoYieldAgent.asset.id,
+          },
+          {
+            name: "Start-Date",
+            value: aoYieldAgent.startDate.toString(),
+          },
+          {
+            name: "End-Date",
+            value: aoYieldAgent.endDate.toString(),
```
Danziger, line 89: When using run indefinitely, the end date will show up like this:

image

Danziger:
@pawanpaudel93 Not sure if it's related to this PR, but I got two different wAR:

image

This is how they appear under settings:

image

Also, I got a (pending) transaction from the agent:

image

But it doesn't show up in the dashboard:

image

Maybe it would be good to make sure all tokens involved in an agent are added to the dashboard, even if they are going to display 0 initially.

Also, the "Cancelled" status is a bit confusing:

image

It seems like I can click it and re-enable it, but that's not the case, I simply get a screen to configure a new one from scratch. If that's the case, I would rename the button to "Remove" instead of "Cancel" and get rid of the "Cancelled" status. Just show "Active" or nothing.

Lastly, would it make sense to add more than one agent (e.g. to convert 50% of my yields to wUSDC and the other 50% to wAW)?

Danziger:
@martonlederer For a fraction of a second, `NaN` is shown here:

image

Danziger:
@pawanpaudel93 @martonlederer I miss a description of what each agent does, for both the yield agent and the LiquidOps agent.

Also, the way we display an active agent is inconsistent between the two:

image

The liquid ops agent takes a couple of seconds to display the active agents. Adding a loading state would be nice. Otherwise, if the agent is clicked before it loads, the user goes to the create agent view, rather then seeing the one they already have.

Lastly, I'm a bit confused with the term agent as the yield agent is doing something regularly, while the LiquidOps ones work more like a widget, where I do something manually and after that they simply show some information, but lack the ability to regularly transfer tokens to LiquidOps (similarly to what the yield agent does).

Danziger:
@martonlederer This seems to be a bug:

image

If I've deposited wUSDC, I should have owUSDC, right?

Danziger:
@martonlederer The balance component should be updated to disable the ability to show/hide its value, as that state is persisted and also affects the balance on the dashboard:

image

Danziger:
@pawanpaudel93 When opening the extension, the agents banner is visible for a fraction of a second even if it has been dismissed already.

src/routes/popup/agents/liquidops/[agent].tsx
Comment (Danziger) [inline src/routes/popup/agents/liquidops/[agent].tsx:1]: Use regular names for files, as we are not using file-based routing.

martonlederer:
> I miss a description of what each agent does, for both the yield agent and the LiquidOps agent.

I have not received any specs or designs for that.

> Lastly, I'm a bit confused with the term agent as the yield agent is doing something regularly, while the LiquidOps ones work more like a widget, where I do something manually and after that they simply show some information, but lack the ability to regularly transfer tokens to LiquidOps (similarly to what the yield agent does).

The wording/terms were something the CommunityLabs team came up with.

> If I've deposited wUSDC, I should have owUSDC, right?

I think you've deposited too little, so it rounds down.

> The balance component should be updated to disable the ability to show/hide its value, as that state is persisted and also affects the balance on the dashboard

I thought the goal of that feature is to disable showing quantities. Happy to remove it tho.

src/routes/popup/agents/liquidops/agents.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Flex } from "~components/common/Flex";
+import { Line } from "~routes/popup/purchase";
+import { Agent } from "../components/liquidops/Agent";
+import { tokenData } from "liquidops";
+import { useActiveTokens } from "./utils/hooks/useAvailableTokens";
+import { useEffect, useMemo } from "react";
+import { trackEvent, EventType, trackPage, PageType } from "~utils/analytics";
+
+export function LiquidOpsAgentsView() {
+  const availableTokens = Object.values(tokenData).filter((token) => !token.deprecated);
```
Danziger: You could use a single `useMemo` that returns `{ availableTokens, inactiveTokens }`

src/routes/popup/agents/liquidops/confirm.tsx
```diff
@@ -0,0 +1,253 @@
...
+        ]);
+      } else {
+        console.warn("Error trying to lend/withdraw: " + (error?.message || error || "Unknown error"));
+      }
+
+      navigate(`/agents/liquidops/${ticker}/${action}/result/${!error ? "success" : "failure"}`);
+    },
+  });
+
+  const activeTokens = Object.values(tokenData).filter((token) => !token.deprecated);
```
Danziger: Maybe you could move this logic inside the `useMemo` below.

src/routes/popup/agents/liquidops/[agent].tsx
Comment (martonlederer) [reply src/routes/popup/agents/liquidops/[agent].tsx:1]: I was following the format of the [token route](https://github.com/wanderwallet/Wander/blob/production/src/routes/popup/token/%5Bid%5D.tsx), but I've also seen it [here](https://github.com/wanderwallet/Wander/tree/production/src/routes/popup/settings/wallets/%5Baddress%5D). Happy to change it tho.

src/routes/popup/agents/liquidops/utils/LiquidOps.ts
```diff
@@ -0,0 +1,51 @@
+import LiquidOps from "liquidops";
+import { ArweaveSigner, createData } from "@dha-team/arbundles";
+import { getActiveKeyfile, type DecryptedWallet } from "~wallets";
+import { isLocalWallet } from "~utils/assertions";
+import { freeDecryptedWallet } from "~wallets/encryption";
+import type { JWKInterface } from "arweave/web/lib/wallet";
+
+export const LiquidOpsClient = async (useWallet = false) => {
+  const decryptedWallet: DecryptedWallet = useWallet
```
Danziger: You might want to wrap this in a `try-catch`. Otherwise, `getActiveKeyfile()` might get called but not `freeDecryptedWallet()`, if there's an error.

src/routes/popup/agents/liquidops/utils/hooks/useAvailableTokens.ts
```diff
@@ -0,0 +1,74 @@
...
+export function useActiveTokens() {
+  const [activeAddress] = useStorage<string>({
+    key: "active_address",
+    instance: ExtensionStorage,
+  });
+
+  return useQuery({
+    queryKey: ["activeTokens", activeAddress],
+    queryFn: async () => {
+      const { client } = await LiquidOpsClient();
```
Danziger, line 32: Not calling `free()` here.

src/routes/popup/agents/liquidops/utils/hooks/useEarnings.ts
```diff
@@ -0,0 +1,57 @@
...
+    profit: BigNumber;
+    startDate?: number;
+  }>({
+    queryKey: ["earnings", ticker, activeAddress],
+    queryFn: async () => {
+      const activeTokens = Object.values(tokenData).filter((token) => !token.deprecated);
+      const token = activeTokens.find((token) => token.ticker.toLowerCase() === ticker.toLowerCase());
+
+      try {
+        const { client } = await LiquidOpsClient();
```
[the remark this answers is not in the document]
martonlederer: The decrypted wallet is not used in this function.

Danziger:
> > I miss a description of what each agent does, for both the yield agent and the LiquidOps agent.
> 
> I have not received any specs or designs for that.
> 
> > Lastly, I'm a bit confused with the term agent as the yield agent is doing something regularly, while the LiquidOps ones work more like a widget, where I do something manually and after that they simply show some information, but lack the ability to regularly transfer tokens to LiquidOps (similarly to what the yield agent does).
> 
> The wording/terms were something the CommunityLabs team came up with.
> 
> > If I've deposited wUSDC, I should have owUSDC, right?
> 
> I think you've deposited too little, so it rounds down.
> 
> > The balance component should be updated to disable the ability to show/hide its value, as that state is persisted and also affects the balance on the dashboard
> 
> I thought the goal of that feature is to disable showing quantities. Happy to remove it tho.

@martonlederer Ok, let me discuss these issues internally. Regarding the rounding down, yep, probably because of that. Wouldn't it make more sense to show the whole quantity when it is below 0.01? Or display it as `>0.01` or `~0`.

src/routes/popup/agents/liquidops/[agent].tsx
Comment (Danziger) [reply src/routes/popup/agents/liquidops/[agent].tsx:1]: Yep, those should be updated as well at some point. I already mentioned this to Pawan: Unless I'm pointing out actual buts, feel free to consider my comments suggestions for improvement, which we can simply create Jira issues to address them later. This would be one of those.

src/wallets/router/router.utils.ts
```diff
@@ -122,6 +122,14 @@ export function useLocation() {
         }
       }
 
+      // Handle path parameters
+      if (options?.params) {
+        Object.entries(options.params).forEach(([key, value]) => {
+          toPath = toPath.replace(`:${key}`, value.toString()) as WanderRoutePath;
```
Danziger, line 131: @pawanpaudel93 You might want to add a check for development only:

```
if (process.env.NODE_ENV === "development" && toPath.indexOf(`:${key}) === -1) throw new Error(`Path ${ to } has no ${ key } param.`);
```

src/routes/popup/agents/components/LiquidOpsAgentListItem.tsx
pawanpaudel93: I will fix this in a new PR.

src/routes/popup/agents/ao-yield/confirm-agent.tsx
pawanpaudel93: Yes, i set the value as a max date timestamp when indefinitely is chosen so we have a value. Should i maybe hide it or show anything when indefinitely is chosen ?

Danziger: Yes, I would just hide it as there's another row that says that it's running indefinitely anyway.

Danziger (approved):

PR #864, opened 2025-06-15
feat: ARC-1271: Connect SDK `signOut()` method

Danziger:
- Add `signOut()` method to Wander SDK. Closes https://communitylabs.atlassian.net/browse/ARC-1271
- Add `setTheme()` method to Wander SDK.
- Reject all pending auth requests on sign out.
- Update how the SDK `Iframe` and `Button` components store and update their CSS variables to also update them when the root/global theme changes. Make sure each change/transitions happen at the right time and speed.
- Fix `Button` hover styles (for all 3 variants).
- Fix theme flicker. Closes https://communitylabs.atlassian.net/browse/ARC-1369
- Fix theme options merge. Closes https://communitylabs.atlassian.net/browse/ARC-1279
- Fix modal moving to the bottom-right corner for a brief moment when it loads.
- Fix modal occasionally resizing horizontally indefinitely. Closes https://communitylabs.atlassian.net/browse/ARC-1325
- Fix SDK button requests counter not resetting on sign out.
- Improve post message handling to prevent errors being thrown when receiving messages from external sources with unexpected formats.
- Fix copy button in wallet dropdown closing the dropdown when clicked. Closes https://communitylabs.atlassian.net/browse/ARC-1281


assets/popup.css
```diff
@@ -157,13 +157,23 @@ button {
...
   z-index: 99999999;
-  background-color: var(--backgroundColor, var(--defaultBackgroundColor, white));
   transition:
     display linear 230ms allow-discrete,
     background-color linear 230ms,
+    color linear 230ms,
     opacity linear 230ms;
 }
 
+html:not(#wanderConnectRoot) #cover {
```
Danziger, line 167: This was causing a small flicker when loading the app, as previously the background color had been fixed for the `html` element, but not for `#cover`.

src/components/embed/contexts/ThemeContext.tsx
```diff
@@ -48,36 +58,50 @@ export const ThemeProvider = ({ children }: { children: ReactNode }) => {
...
+      // If mode is (`mode` arg) or was (`lastModeRef.current`) "system" and the
+      // applied theme doesn't change, no need to show the cover, just update the state.
+      setMode(mode);
+
+      return;
     }
-  }, [mode, isDarkMode]);
+
+    window.clearTimeout(setModeTimeoutIDRef.current);
+    cover.removeAttribute("aria-hidden");
```
Danziger, line 92: Because some elements have transitions and some other ones don't, changing the theme with the iframe open looks weird, as some colors will change instantly while some other ones will transition. To address that, we should the spinner first and hide it again shortly after (once the transitions are done).

src/contents/api.ts
```diff
@@ -36,9 +36,7 @@ container.removeChild(script);
 
 window.addEventListener("message", async ({ data }: MessageEvent<ApiCall>) => {
   // verify that the call is meant for the extension
-  if (data.app !== "wander") {
```
Danziger, line 39: This will throw an error if the dApp posts a message with non-object data.

src/utils/auth/auth.utils.ts
```diff
@@ -207,39 +207,47 @@ export async function createAuthPopup(authRequestData: null | AuthRequestData, m
...
 type AuthResultCallback<T> = (data: T) => void;
 
 const authResultCallbacks = new Map<string, AuthResultCallback<any>>();
 
-function addAuthResultListener<T>(authID: string, popupWindowTabID: number, fn: AuthResultCallback<T>) {
+function addAuthResultListener<T>(authID: string, fn: AuthResultCallback<T>) {
   authResultCallbacks.set(authID, fn);
 
-  if (authResultCallbacks.size === 1) {
-    isomorphicOnMessage("auth_result", ({ sender, data }) => {
```
Danziger, line 218: This would call `isomorphicOnMessage` more than once, as `authResultCallbacks` is constantly growing and shrinking, so `authResultCallbacks.size === 1` will be `true` multiple times.

wander-connect-sdk/src/wander-connect.ts
```diff
@@ -544,6 +551,51 @@ export class WanderConnect {
...
+      const contentWindow = iframeRef?.contentWindow;
+
+      if (!iframeRef || !contentWindow) {
+        reject(new Error("Missing Wander Connect iframe"));
+        return;
+      }
+
+      isomorphicSendMessage({
+        destination: "background",
+        // @ts-ignore
```
Danziger, line 577: `shim.d.ts` doesn't work inside `wander-connect-sdk`, so types for `isomorphicSendMessage()` are missing.

src/components/ResizeEventObserver.tsx
```diff
@@ -22,6 +21,10 @@ export function ResizeEventObserver({ containerRef }: ResizeEventObserver) {
 
     if (!containerElement || import.meta.env?.VITE_IS_EMBEDDED_APP !== "1") return;
 
+    const location = locationRef.current;
+
+    if (location.startsWith("/__OVERRIDES/")) return;
```
Danziger, line 26: While the SDK would default to a "modal" layout when it first loads, the app was sending a "popup" layout event on load (when `location === "/__OVERRIDES/cover"`. That would cause the iframe to move to the bottom-right corner for a moment.

src/components/embed/contexts/ThemeContext.tsx
```diff
@@ -48,36 +57,52 @@ export const ThemeProvider = ({ children }: { children: ReactNode }) => {
...
+    setModeTimeoutIDRef.current = window.setTimeout(() => {
+      setMode(mode);
+      setModeTimeoutIDRef.current = window.setTimeout(() => {
+        cover.setAttribute("aria-hidden", "true");
+      }, 230);
+    }, 230);
```
Danziger, line 98: The fist timeout waits for the `#cover` element to appear, while the second one waits for the theme to change.

wander-connect-sdk/src/utils/styles/styles.utils.ts
```diff
@@ -37,9 +61,6 @@ export function mergeCSSVariablesOption<T extends Object>(
     }
   }
 
-  if (themeOption === "light") cssVarsDark = cssVarsLight;
```
Danziger, line 40: Before, we would set both the light and dark CSS variables to the same value if theme was set to `"light"` or `"dark"`, but this prevents us from changing the theme later, as both CSS variable sets are the same.

Now we store each set (light and dark) regardless or the theme option, and apply the same one twice (e.g. use the light CSS variables for the light variables definitions and for the dark ones too) if needed, so that we can change the them anytime.

wander-connect-sdk/src/wander-connect.ts
```diff
@@ -544,6 +558,64 @@ export class WanderConnect {
...
+      // @ts-ignore
+      data: theme,
+    });
+
+    window.clearTimeout(this.setThemeTimeoutID);
+
+    this.setThemeTimeoutID = window.setTimeout(() => {
+      if (!this.hasIndependentIframeTheme && this.iframeComponent) this.iframeComponent.setTheme(theme);
+      if (!this.hasIndependentButtonTheme && this.buttonComponent) this.buttonComponent.setTheme(theme);
+    }, 230);
```
Danziger, line 603: Waits for the `#cover` element to appear.

wander-connect-sdk/src/wander-connect.ts
```diff
@@ -544,6 +558,64 @@ export class WanderConnect {
...
+      // @ts-ignore
+      messageId: "embedded_setTheme",
+      // @ts-ignore
+      data: theme,
+    });
+
+    window.clearTimeout(this.setThemeTimeoutID);
+
+    this.setThemeTimeoutID = window.setTimeout(() => {
+      if (!this.hasIndependentIframeTheme && this.iframeComponent) this.iframeComponent.setTheme(theme);
```
Danziger, line 601: Calling this function changes the app theme as well as the iframe and button theme, as long as those haven't been manually set using their respective and more specific `theme` option.

pawanpaudel93:
@Danziger Even after signOut the pending auth requests are not cancelled and are shown again when signed in again.

wander-connect-sdk/src/wander-connect.ts
```diff
@@ -544,6 +557,64 @@ export class WanderConnect {
...
+        messageId: "embedded_signOut",
+        // @ts-ignore
+        data: undefined,
+      });
+
+      this.signOutMethodCallback = resolve;
+    });
+  }
+
+  public setTheme(theme: ThemeSetting): void {
```
pawanpaudel93, line 587: - I think this function should check if the passed theme is light, dark or system as I passed random text and I could see it being set in localStorage.

- Don't know if it's an issue but I see this error everytime as the passed value is not an object. I saw your comment about throwing error when data is not an object.
image

pawanpaudel93:
- With the config as in the image, even if the iframe theme is set as `dark`, the iframe theme is shown as `light`. `iframe.theme` doesnot work but `button.theme` works fine.
- With the same config as in the image, if the `iframe.theme` and `button.theme` is set and if we use `setTheme` to set any theme like `setTheme("light")`, only iframe theme changes but not button theme.
-  With the same config as in the image, if both `iframe.theme` and `button.theme` is removed then `setTheme` works fine to set the theme for iframe and button.

image

wander-connect-sdk/src/wander-connect.ts
Danziger: Regarding the second error, that's coming from your Wander BE, in the `message` event listener inside `api.ts`. It has been resolved in this PR too:

```
// verify that the call is meant for the extension
if (typeof data !== "object" || data.app !== "wander") return;
```

If you were to build the BE from this branch and replace the current one you are using with that, you should not see this error, unless it's coming from a different extension. In my case, I was getting a similar one from Rainbow, which I reported here: https://github.com/rainbow-me/browser-extension/issues/1904

Danziger: Added validation for the values.

Danziger:
> * With the config as in the image, even if the iframe theme is set as `dark`, the iframe theme is shown as `light`. `iframe.theme` doesnot work but `button.theme` works fine.
> * With the same config as in the image, if the `iframe.theme` and `button.theme` is set and if we use `setTheme` to set any theme like `setTheme("light")`, only iframe theme changes but not button theme.
> * With the same config as in the image, if both `iframe.theme` and `button.theme` is removed then `setTheme` works fine to set the theme for iframe and button.
> 
> image

@pawanpaudel93 This is actually working as intended:

- If you only set the root `theme`, which is what most users will probably do, then the button, the iframe (wrapper) and the app inside the frame all use the same theme.
- If you also set `iframe.theme` or `button.theme`, then `setTheme()` won't change those ones. Instead, you should use `setIframeTheme()` or `setButtonTheme()`.

Also note that in the first issue you mention, `iframe.theme` only affects the iframe HTML element and its wrapper, not the contents of the iframe. It's actually working fine (you can see the border around the iframe is black instead of light grey).

PR #866, opened 2025-06-17
fixes and improvements to ao agents

pawanpaudel93:
## Summary

This PR includes various fixes and enhancements for AO agents, improving performance, UX, and data handling:

### UI/UX Improvements
- **Slider performance**: Improved responsiveness for a smoother user experience.
- **Asset update**: Replaced `war.svg` with `war.png` since the original SVG was a raster image.
- **Slippage modal**:  
  - Buttons now support press-and-hold to increment/decrement slippage values.  
- **Date selector enhancements**:
  - Clicking the end date button now opens the selector with the end date pre-selected.
  - If "Run indefinitely" is checked, the end date button is disabled and no longer opens the selector.
  - Refactored related logic into date utility functions.
  - Added month/year selectors for easier navigation.

### Agent Creation Workflow
- Fields are now persisted when navigating back from the confirmation screen—no need to re-enter previously filled data.
- When "Run indefinitely" is selected, the end date on the confirmation screen now displays as `∞`.

### Token & Notification Enhancements
- **LiquidOps tokens** are now auto-detected and properly reflected in both notifications and transaction history. _Closes [ARC-1380](https://communitylabs.atlassian.net/browse/ARC-1380)_
- Updated balance fetch logic to use the new CU for **ARIO** and **USDA** tokens.  _Closes [ARC-1356](https://communitylabs.atlassian.net/browse/ARC-1356)_

[ARC-1380]: https://communitylabs.atlassian.net/browse/ARC-1380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/popup/agents/components/ao-yield/SlippageSelectorModal.tsx
```diff
@@ -14,75 +15,185 @@ interface SlippageSelectorModalProps {
...
-    if (newValue > 10) {
-      showToast(toasts, false);
-      return;
+    animationFrameRef.current = requestAnimationFrame(updateHold);
+  }, [isHolding, holdDirection, handleIncreaseSlippage, handleDecreaseSlippage]);
+
+  useEffect(() => {
+    if (isHolding && holdDirection) {
+      updateSpeed.current = 0.1;
+      animationFrameRef.current = requestAnimationFrame(updateHold);
```
Danziger: This seems to be too fast, specially considering the values users will be setting will be relatively low. There's a library in the project called `@swyg/corre` that has a `useInterval()` hook. It would probably be easier to implement using that. You just need to set a state with the increment/step and then have your logic to regularly increase/decrease inside this hook. Setting and clearing the interval is managed internally.

It also includes a `useThrottledCallback()` hook you could use for the range input.

Danziger:
The date picker jumps to the selected date when clicking it from outside, but not from inside: 

image

Danziger (approved):

PR #871, opened 2025-06-18
fix: resolve agents routing issues

pawanpaudel93:
## Summary

Resolve routing issues in both ao and liquidops agents routing in order to prevent going back to success/failure screen after create/withdraw/deposit.

### DIFF
--- a/src/routes/popup/agents/ao-yield/agent-activated.tsx
+++ b/src/routes/popup/agents/ao-yield/agent-activated.tsx
@@ -10,7 +10,7 @@ import { useEffect } from "react";
 import { trackPage, PageType } from "~utils/analytics";
 
 export function AOYieldAgentActivatedView() {
-  const { navigate } = useLocation();
+  const { navigate, back } = useLocation();
   const { activationStatus = "success" } = useSearchParams<{ activationStatus: "success" | "error" }>();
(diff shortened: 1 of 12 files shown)


src/routes/popup/agents/ao-yield/agent-activated.tsx
```diff
@@ -10,15 +10,17 @@ import { useEffect } from "react";
...
+  const { navigate, back } = useLocation();
   const { activationStatus = "success" } = useSearchParams<{ activationStatus: "success" | "error" }>();
   const isActivated = activationStatus === "success";
 
   const handlePrimaryAction = () => {
     if (isActivated) {
       navigate(PopupPaths.Home);
     } else {
-      navigate(PopupPaths.CreateAOYieldAgent);
+      // double back to get to the create agent page
```
Danziger, line 21: We should avoid using `back()`, specially like this, as `customHistory` doesn't properly keep track of all pages, so we can't be sure this is going to work as intended. Same comment for `previousLocation`.

In any case, if we need to release something and you've tested it, we can merge this as it is, but keep in mind this should not be a permanent solution and we should avoid using `back` /  `customHistory` / `previousLocation`, at least until the underlaying issue is fixed.

PR #875, opened 2025-06-19
fix: analytics events & transaction details fixes

pawanpaudel93:
## Summary

This PR addresses multiple issues related to analytics and transaction display:

* Fixes analytics events not being sent as expected.
* Updates the transaction details screen to:
  * Correctly display LiquidOps mint transactions.
  * Accurately show the sending address when the AO Authority initiates the message.
* Fixes a bug where agent cancellation was not working properly.

### DIFF
--- a/src/routes/popup/transaction/[id].tsx
+++ b/src/routes/popup/transaction/[id].tsx
@@ -148,7 +148,7 @@ export function TransactionView({ params: { id, gateway: gw, message } }: Transa
   useEffect(() => {
     if (!id || !graphqlGateways.length) return;
 
-    let timeoutID: NodeJS.Timeout | undefined;
+    let timeoutID: number | undefined;
     let fetchCount = 0;
 
     const fetchTx = async () => {
@@ -220,7 +220,7 @@ export function TransactionView({ params: { id, gateway: gw, message } }: Transa
 
       if (!data.transaction) {
         fetchCount++;
-        timeoutID = setTimeout(fetchTx, 5000);
+        timeoutID = window.setTimeout(fetchTx, 5000);
       } else {
         timeoutID = undefined;
         try {
--- a/src/routes/popup/transaction/[id].tsx
+++ b/src/routes/popup/transaction/[id].tsx
@@ -82,17 +82,25 @@ export function TransactionView({ params: { id, gateway: gw, message } }: Transa
     instance: ExtensionStorage,
   });
 
-  const ownerAddress = transaction?.owner.address;
-  const fromAddress = useMemo(
-    () =>
-      ao.isAo && ownerAddress === AO_AUTHORITY_ID
+  const { isAo } = ao;
+
+  const { fromAddress, toAddress, fromMe, toMe } = useMemo(() => {
+    const ownerAddress = transaction?.owner.address;
+    const fromAddress =
+      isAo && ownerAddress === AO_AUTHORITY_ID
         ? getTagValue("From-Process", transaction?.tags || []) || ownerAddress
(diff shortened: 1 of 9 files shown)


src/routes/popup/transaction/[id].tsx
```diff
@@ -141,7 +148,7 @@ export function TransactionView({ params: { id, gateway: gw, message } }: Transa
   useEffect(() => {
     if (!id || !graphqlGateways.length) return;
 
-    let timeoutID: number | undefined;
+    let timeoutID: NodeJS.Timeout | undefined;
```
Comment (Danziger) [inline src/routes/popup/transaction/[id].tsx]: This is `number`. If you set it with `timeoutID = window.setTimeout` instead of `timeoutId = setTimeout` you'll have no issue with the types.

src/routes/popup/transaction/[id].tsx
```diff
@@ -80,17 +82,22 @@ export function TransactionView({ params: { id, gateway: gw, message } }: Transa
     instance: ExtensionStorage,
   });
 
-  const fromAddress = transaction?.owner.address;
+  const ownerAddress = transaction?.owner.address;
```
Comment (Danziger) [inline src/routes/popup/transaction/[id].tsx]: It looks like every time the transaction changes, this line and all 3 memos will run, so it's better to use a single `useMemo`:

```
const { isAo } = ao;

const {
  fromAddress,
  toAddress,
  fromMe,
  toMe,
} = useMemo(() => {
  const ownerAddress = transaction?.owner.address;
  const fromAddress = isAo && ownerAddress === AO_AUTHORITY_ID
    ? getTagValue("From-Process", transaction?.tags || []) || ownerAddress
    : ownerAddress;
  const toAddress = transaction?.recipient;
  const fromMe = wallets.find((wallet) => wallet.address === fromAddress);
  const toMe = wallets.find((wallet) => wallet.address === toAddress);

  return {
    fromAddress,
    toAddress,
    fromMe,
    toMe,    
  };
}, [isAo, transaction]);
```

If this functionality happens in other places, this is also a good candidate to be extracted as a custom hook or as a separate util function (non-memoized, each view can wrap it in `useMemo` if needed).

src/utils/agents/swap.ts
```diff
@@ -671,8 +670,8 @@ export async function checkIfRecentTxSwapSucceeded(): Promise<boolean> {
 
     const foundTxIds = new Set<string>();
 
-    try {
-      for (const edge of edges) {
+    for (const edge of edges) {
```
Danziger: Better to run them in parallel:

```
const trackDirectPromises = edges.map((edge) => {
        const tags = edge.node.tags;
        const txId = getTagValue("Pushed-For", tags);
        const buyAsset = getTagValue("Token-Out", tags);
        const sellAmount = getTagValue("Amount-In", tags);
        const buyAmount = getTagValue("Amount-Out", tags);
        const dexUsed = getTagValue("Dex", tags);
        const wanderFee = getTagValue("Swap-Fee", tags);

        if (foundTxIds.has(txId)) continue;

        foundTxIds.add(txId);

        return trackDirect(EventType.AO_YIELD_AGENT_TRANSACTION, {
          buyAsset,
          sellAmount,
          buyAmount,
          dexUsed,
          wanderFee,
        }).catch((err) => {
          log(LOG_GROUP.AGENTS, "Error tracking recent tx: ", err);
          throw err;
        });
});

Promise.allSettled(trackDirectPromises);
```

Note I've replaced the `try-catch` with a `.catch()` because I assume if there's an error, it's going to come from `trackDirect`, not from the `getTagValue()` calls above.

Danziger (approved):

PR #876, opened 2025-06-19
feat: use ar.io sdk for ArNS operations

kunstmusik:
Adds ar.io SDK and replaces custom lookup code and types for ArNS with ones from SDK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     ]
   },
   "dependencies": {
+    "@ar.io/sdk": "^3.13.0",
     "@arconnect/components": "^1.0.0",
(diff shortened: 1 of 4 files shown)


Danziger:
@kunstmusik Thanks for the PR. Code looks good but we want to look into an issue @pawanpaudel93 found with the bundle size growing ~10mb. This is probably not on your side. We suspect it's either a bug on Plasmo or a result of our specific build setup. I'll look into this ASAP, but most likely it will be at the end of next week. Depending on we find, we'll fix the tree shaking issue and merge your PR, or let you know if we think the issue might be cause by your library.

7i7o:
Merged in #948

PR #890, opened 2025-06-27
feat: ARC-1400: Wallet activation and rotation retry

Danziger:
This PR also fixes an issue when requesting unpartitioned state, that was causing some values to be read/written into the regular `localStorage` instead. Additionally:

- Re-attempt wallet activation 3 times. Closes https://communitylabs.atlassian.net/browse/ARC-1327
- Fix wallet activation/rotation issues. Closes https://communitylabs.atlassian.net/browse/ARC-1400
- Fix inability to reactive a wallet after clearing browser data. Closes https://communitylabs.atlassian.net/browse/ARC-1263

Backend PR: https://github.com/wanderwallet/embed-api/pull/39 (closes https://communitylabs.atlassian.net/browse/ARC-1406, even thought the root cause was still the issues described above).

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "copy-to-clipboard": "^3.3.2",
     "currency-symbol-map": "^5.1.0",
     "dayjs": "^1.11.6",
-    "embed-api": "https://github.com/wanderwallet/embed-api#feat/arc-1400-fix-challenges",
+    "embed-api": "https://github.com/wanderwallet/embed-api#ec23a9586004c3742a3edb512bf753786d9dcdf5",
     "framer-motion": "^11.11.7",
     "human-crypto-keys": "^0.1.4",
     "js-confetti": "^0.12.0",
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "copy-to-clipboard": "^3.3.2",
     "currency-symbol-map": "^5.1.0",
     "dayjs": "^1.11.6",
-    "embed-api": "https://github.com/wanderwallet/embed-api#5dbcb8722252cb6a0a7ca6092d93b18d4ac071d0",
+    "embed-api": "https://github.com/wanderwallet/embed-api#feat/arc-1400-fix-challenges",
     "framer-motion": "^11.11.7",
     "human-crypto-keys": "^0.1.4",
     "js-confetti": "^0.12.0",
--- a/src/iframe/storage/unpartitioned-storage/local-storage.ts
+++ b/src/iframe/storage/unpartitioned-storage/local-storage.ts
@@ -1,3 +1,7 @@
+import {
+  getUnpartitionedStateStatus,
+  HAS_SIMPLE_STORAGE_API,
+} from "~iframe/storage/unpartitioned-storage/unpartitioned-storage.utils";
 import { EnhancedStorage } from "./unpartitioned-storage";
 
 export class LocalStorage extends EnhancedStorage {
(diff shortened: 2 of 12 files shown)


src/utils/embedded/device-nonce/device-nonce.utils.ts
```diff
@@ -36,6 +36,8 @@ export function generateDeviceNonce(): DeviceNonce {
 }
 
 export async function storeDeviceNonce(deviceNonce: DeviceNonce): Promise<DeviceNonce> {
+  console.trace("Updating deviceNonce =", deviceNonce);
```
Danziger, line 39: When playing around with Odysee and the Playground on different tabs, I sometimes get different deviceNonce values, so I want to see if it's because we are overriding it for some reason.

src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -8,10 +12,18 @@ export class LocalStorage extends EnhancedStorage {
   }
 
   static async getInstance(): Promise<LocalStorage> {
-    if (!this.instance) {
-      this.instance = new LocalStorage();
-      await this.instance.requestStorageAccess();
```
Danziger, line 13: I think this is the main cause of the share recovery issue.

Because `getInstance()` is called from different places when the app just loads, only the first call will `await this.instance.requestStorageAccess()`. The other ones will return `this.instance` straight away, before the `requestStorageAccess` call has resolved, so `instance.storage` in those cases would be `localStorage` still, instead of `handle.localStorage`. That mean that the `DEVICE_NONCE` would be set in the regular, partitioned, `localStorage`, not on the unpartitioned one. Because of that, if you open Wander Connect in 2 different origins, each of them will get a different device nonce, but the Supabase session will be shared, so they'll be competing to try to update the session with their device nonce. This means that sometimes a challenge will be validated just fine, if the session was already updated, and sometimes it won't, if the session still included the device nonce from the other domain.

This is also was was causing some challenges to be updated (upserted) when reloading, despite the challenge solution generation incorrectly using `createdAt` instead of `updatedAt` (which should not affect the happy path).

I'll open a PR on the backend next, but this one can already be merged.

package.json
```diff
@@ -85,7 +85,7 @@
     "copy-to-clipboard": "^3.3.2",
     "currency-symbol-map": "^5.1.0",
     "dayjs": "^1.11.6",
-    "embed-api": "https://github.com/wanderwallet/embed-api#5dbcb8722252cb6a0a7ca6092d93b18d4ac071d0",
+    "embed-api": "https://github.com/wanderwallet/embed-api#feat/arc-1400-fix-challenges",
```
Danziger: Merge backend first, then change this before merging.

src/iframe/storage/unpartitioned-storage/local-storage.ts
pawanpaudel93: @Danziger Since we call `requestStorageAccess` for every `getInstance` call now, once we get the handle should we cache it and reuse or request handle everytime ?

Danziger: It could be an improvement but I don't think we need it now, as we need to merge this fix as soon as possible.

While calling `requestStorageAccess` twice results in 2 different `handle` and `handle.localStorage`, both give access to the exact data, so it's not a big issue.

pawanpaudel93 (approved):
tACK

PR #891, opened 2025-06-27
feat: ARC-1332: Allow creating a new wallet if recovery methods do not exist or are lost

Danziger:
Wallet recovery screen 1) with backups 2) with no backups 3) with no backups in a browser without unpartitioned state, and 4) wallet creation confirmation:

Add / create wallet screens with updated messages on the recovery flow:

Closes https://communitylabs.atlassian.net/browse/ARC-1332.


7i7o (approved):
tACK 👍

PR #898, opened 2025-07-01
feat: ARC-1198: Hash based challenges flag

Danziger:
Add a flag to skip the generation of the private key, once the backend switches to hash-based challenges.

Note the public keys are still generated, as not generating them means we cannot easily go back to the signature-based system, because the wallets in the DB won't have a public key stored. Later on, with some changes on the backend as well, we can update the system so that switching from one method to the other is possible at any point, even with public key generation is also disabled and some wallets end up missing their public key.

Note currently `SHARE_ACTIVE_TTL_MS` = 10 minutes, so the public key still needs to be generated relatively often. Changing that value to a few hours will also help speed things up.

Eventually, it could make sense to allow users to tweak these and some other security-related options to balance speed, convenience and security.

Closes https://communitylabs.atlassian.net/browse/ARC-1198.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,9 @@ VITE_PROD_EMBEDDED_SERVER_BASE_URL="https://connect-api.wander.app"
 VITE_TEST_DEFAULT_EMBEDDED_CLIENT_ID="test"
 VITE_TEST_DEFAULT_EMBEDDED_SERVER_BASE_URL="https://test.com"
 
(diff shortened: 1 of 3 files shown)


7i7o (approved):
tACK 👍

Danziger:
Warning: Only merge once the backend ENV variable has been updated.

Danziger:
The version with only hashes is not good in terms of security, so I'll be looking into other signature algorithms, most likeyly Ed25519 (+ HKDF).

PR #899, opened 2025-07-02
feat: ARC-1294: OTP authentication flows.

Danziger:
Closes https://communitylabs.atlassian.net/browse/ARC-1294.

**Screenshots:**


src/routes/embedded/auth/auth-email-otp/auth-email-otp.view.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+      if (otpCode.length !== OTP_LENGTH) {
+        toast.error(`Please enter all ${OTP_LENGTH} digits of the access code`);
+        return;
+      }
+
+      setIsVerifying(true);
+
+      try {
+        await authenticate({
```
Danziger, line 117: This now takes care of both sign ups and OTP sign ins, so the sign up view is gone.

src/routes/embedded/auth/auth-email-verify/auth-email-verify.view.tsx
```diff
@@ -9,24 +9,29 @@ import { OnboardingCard } from "~components/embed/ui/molecules/card/onboarding-c
...
   OPT_COOLDOWN_DURATION_SEC,
   OTP_LENGTH,
-  type CodeInputHandle,
-} from "~components/embed/ui/atoms/code-input/CodeInput";
-import { useCooldownCallback } from "~utils/react/useCooldownCallback";
+  setOtpAvailable,
+} from "~utils/otp/otp.utils";
+import { useAsyncEffect } from "~utils/react/useAsyncEffect";
 
 export function AuthEmailVerifyEmbeddedView() {
```
Danziger, line 23: This view is only here to handle a very unlikely error when signing in with password to a non-verified account.

src/routes/embedded/auth/auth/auth.view.tsx
```diff
@@ -105,9 +113,14 @@ export function AuthEmbeddedView() {
         return;
       }
 
-      navigate(isAlreadyRegistered ? EmbeddedPaths.AuthEmailSignin : EmbeddedPaths.AuthEmailSignup, {
-        search: { email },
-      });
+      navigate(
+        isAlreadyRegistered || preferredEmailAuth !== "password"
```
Danziger: Users who used email and password will keep seeing the password screen instead of the OTP one if they come back to the same device.

src/utils/authentication/authentication.service.ts
```diff
@@ -195,16 +195,6 @@ async function verifyOtp(authParams: AuthVerifyOtpParams) {
 
   if (error) throw error;
 
-  if (!(data.user.user_metadata as SupabaseUserMetadata).hasPassword) {
```
Danziger, line 198: Before, any sign up was created with password, but not anymore. This same `updateUser` call is still performed when setting/changing a password from Actions.

src/utils/otp/otp.utils.ts
```diff
@@ -0,0 +1,38 @@
+import { PersistentStorage } from "~utils/storage";
+import { StorageKeys } from "~utils/storage/storage.constants";
+
+export const OTP_LENGTH = 6;
+export const OPT_COOLDOWN_DURATION_SEC = 60;
+export const OTP_EXPIRATION_SEC = 86400;
```
Danziger: @7i7o Maybe we should limit this to 1h or so?

7i7o: Yes, agree 1 hour should suffice since that's also a lot for a OTP.

Also: there's a typo in the constant name in line 5: OPT instead of OTP

src/routes/embedded/auth/auth/auth.view.tsx
7i7o: "Use password instead" takes me to a sign in page even though I've never set up my password
For a new user and also for a user that has already used an OTP before
This is getting me this "Invalid Login Credentials" error since I obviously don't have a password stored in the server

Screenshot 2025-07-03 at 12 43 01 PM

7i7o: Fixed 👍

7i7o (approved):
tACK 👍

PR #900, opened 2025-07-03
fix: Fix BE build and add some Firefox-specific fixes

Danziger:
This doesn't make the extension work on Firefox but it fixes some errors I saw there.

For it to work on Firefox, we need to see why, once the auth popup opens, it doesn't receive the messages.


7i7o (approved):
tACK 👍

7i7o (changes requested):
This PR breaks the Menu screen

7i7o (approved):
tACK 👍

PR #902, opened 2025-07-03
feat: ARC-1331: Shorter onboarding

Danziger:
- The wallet/account confirmation screen has been removed.
- The backup reminder screen has been replaced with an inline snackbar in the wallet home.
- The snackbar is non-dismissible, so the `skipBackUp` method has been removed from the provider, and the `doNotAskAgainForBackup` method has been removed from `WalletService`. Related backend PR (can be merged after this one): https://github.com/wanderwallet/embed-api/pull/43 
- The snackbar shows as a warning (yellow), except in browsers that do not support unpartitioned state, where it is displayed as an error (red).
- Note the color of the buttons and links inside the snackbar should probably be different, but that will be addressed here in https://communitylabs.atlassian.net/browse/ARC-1315.
- Closes https://communitylabs.atlassian.net/browse/ARC-1331.

Some things initially listed in ARC-1331 have not been implemented:

- The wallet selector doesn't show a warning icon next to each wallet that needs to be backed up. Once we add support for multiple wallets, that can be added.
- The warning is non-dismissible no matter what.
- It doesn't change from warning to error if the user has funds, as checking the total balance is not enough. This can be implemented later if needed.

**Screenshots:**


7i7o (approved):
tACK 👍 

Minor glitch: Copy Seedphrase button is not working

Danziger:
> tACK 👍
> 
> Minor glitch: Copy Seedphrase button is not working

Fixed in https://github.com/wanderwallet/Wander/pull/899

PR #908, opened 2025-07-07
fix: Allow non email users to use email

Danziger:
Before OTP, the `user_exists_by_email` would only check whether a user linked to a given email exists, regardless of what authentication method it has used. We then updated that function to check whether an user with the Email provider exists for that email.

Now we actually want the old behavior, because any existing user can use OTP to authenticate regardless of whether they were created using OAuth.

This a quick client fix to get the old behavior, but the backend will be updated to use the old version of the function too.


7i7o (approved):
tACK 👍

PR #909, opened 2025-07-08
fix: WNDR token progress bar improvements

Danziger:
It also fixes some incorrect React Query's `queryClient` import (if we use different client instances, we'll also have different caches, and invalidating a key on one instance won't invalidate the cache entry on the other one).

image

### DIFF
--- a/src/components/popup/tier/ProgressBar.tsx
+++ b/src/components/popup/tier/ProgressBar.tsx
@@ -1,12 +1,25 @@
 import { Text } from "@arconnect/components-rebrand";
-import { useMemo } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { useCallback, useMemo } from "react";
 import styled from "styled-components";
+import { ExtensionStorage } from "~utils/storage";
+import type { ActiveTier, Tier } from "~utils/tier/types";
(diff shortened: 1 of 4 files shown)


src/components/popup/tier/ProgressBar.tsx
```diff
@@ -43,6 +56,37 @@ export function ProgressBar({
...
+  const activeAddress = useActiveAddress();
+
+  const handleSegmentClicked = useCallback(
+    async (segmentData: ProgressSegmentWithData) => {
+      if (process.env.NODE_ENV !== "development") return;
+
+      const index = segments.findIndex(({ name }) => name === segmentData.name);
+      const rank = segments.length - index;
+      const progress = segmentData.startPercentage + (segmentData.endPercentage - segmentData.startPercentage) / 2;
+      const balance = `${10 ** (18 + index)}`;
```
Danziger, line 70: The fake balance will increase 10x on each tier.

7i7o (approved):
tACK 👍 

Although the `SegmentLabel`s were not center aligned in the designs

pawanpaudel93 (approved):
tACK 🚀

PR #915, opened 2025-07-09
fix: issue where id wasn't being passed to bg

nicholaswma:


Danziger (approved):
utACK, it looks like we've never defined a function without a foreground module, even though the code seems to support that for cases like this where we simply want to forward the params:

Image

Just a comment, it doesn't make sense to update this now.

PR #916, opened 2025-07-09
feat: ARC-1198: EdDSA signatures

Danziger:
Went for https://github.com/paulmillr/noble-curves instead of https://github.com/paulmillr/noble-ed25519 because in the latter it mentions:

> The module is a sister project of [noble-curves](https://github.com/paulmillr/noble-curves), focusing on smaller attack surface & better auditability. Curves are drop-in replacement and have more features: Common.js, ristretto255, X25519, curve25519, ed25519ph. To upgrade from v1 to v2, see [Upgrading](https://github.com/paulmillr/noble-ed25519#upgrading).

And

> noble-ed25519 v2 features improved security and smaller attack surface. The goal of v2 is to provide minimum possible JS library which is safe and fast.
> 
> That means the library was reduced 4x, to just over 300 lines. In order to achieve the goal, some features were moved to [noble-curves](https://github.com/paulmillr/noble-curves), which is even safer and faster drop-in replacement library with same API.

And also:

> The module is production-ready.
>
> We cross-test against sister project [noble-curves](https://github.com/paulmillr/noble-curves), which was audited and provides improved security.
> 
> The current version has not been independently audited. It is a rewrite of v1, which has been audited by cure53 in Feb 2022: [PDF](https://cure53.de/pentest-report_ed25519.pdf).
> It's being fuzzed [in a separate repository](https://github.com/paulmillr/fuzzing)
> If you see anything unusual: investigate and report.

So it looks like the library undergone a massive rewrite / simplification (v1 => v2) and hasn't been audited again, while `noble-curves` has.

Information about audits on Noble: https://paulmillr.com/noble/

Worth keeping an eye on: https://github.com/MystenLabs/ed25519-unsafe-libs


src/utils/embedded/embedded.provider.tsx
```diff
@@ -876,10 +876,11 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
           await WalletUtils.storeEncryptedSeedPhrase(walletId, seedPhrase, jwk).catch(() => {});
         }
         isRecoveryJSON = false;
+        privateKey = jwk;
       } else if (WalletUtils.isRecoveryJSON(recoveryData)) {
-        ({ walletId, recoveryBackupShare, recoveryFileServerSignature } = recoveryData as RecoveryJSON);
-        ({ shareHash: recoveryBackupShareHash, sharePrivateKeyJWK: recoveryBackupSharePrivateKeyJWK } =
-          await WalletUtils.generateShareHashAndPrivateKey(recoveryBackupShare));
+        ({ walletId, recoveryBackupShare, recoveryFileServerSignature } = recoveryData);
```
Danziger, line 825: It would be good to refactor this `if-else` block here so that it's a bit easier to read.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -893,16 +894,25 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
 
       const latestSession = await getLatestSession(session);
 
-      const { shareRecoveryChallenge } = await WalletService.generateWalletRecoveryChallenge({ walletId });
+      const { shareRecoveryChallenge } = await WalletService.generateWalletRecoveryChallenge({
+        walletId,
+        recoveryBackupShareHash,
```
Danziger, line 841: Before, only the `walletId` was needed, but now we need to know what backup share the user is going to use, as old ones should still use RSA challenges, while new ones can use EdDSA challenges.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -893,16 +894,25 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+      });
 
-      const challengeSolution = await ChallengeClientV1.solveChallenge({
+      if (shareRecoveryChallenge.version === "v1") {
+        const derivedRSAKeys = await WalletUtils.deriveRSAKeys(recoveryBackupShare);
```
Danziger, line 845: We only do this when needed, as it's slow. This is fully controlled by the backend, that returns a v1 (RSA) or v2 (EdDSA) challenge depending on the format of the public key that it will use to verify it.

A side-effect of this is that we start deriving the RSA keys a bit later (once we received the challenge, rather than before that, as soon as we start a challenge request-resolve flow), so we might need to extend the challenge TLL. It would probably be good to have different TTLs for RSA and EdDSA.

src/utils/wallets/wallets.utils.ts
```diff
@@ -159,75 +160,68 @@ async function generateShareHash(share: string): Promise<string> {
...
+  log(LOG_GROUP.WALLET_GENERATION, "generateShareHashAndEdKeys()");
 
-  if (!share) {
-    throw new Error("Share is missing — unable to generate share hash and public key.");
-  }
+  const encoder = new TextEncoder();
+  const ikm = encoder.encode(share);
+  // TODO: Change to something else
+  // const salt = crypto.getRandomValues(new Uint8Array(16)); // TODO: Need to be made not random
+  const salt = new Uint8Array(16).fill(0); // For testing purposes, use a fixed salt
```
Danziger: If `share = deviceShare`, we could use `deviceNonce` + `userId` as salt. In this case, if the `deviceNonce` is lost/updated, the wallet will have to be recovered, even if the `deviceShare` is still there.

If `share = recoveryBackupShare`, we could use `userId` as salt, as in this case this needs to be the same across different devices.

src/utils/wallets/wallets.utils.ts
```diff
@@ -159,75 +160,68 @@ async function generateShareHash(share: string): Promise<string> {
...
+  // const info = encoder.encode("context");
+  const info = encoder.encode("");
```
Danziger, line 180: TODO: Fill this in.

package.json
```diff
@@ -86,7 +87,7 @@
     "copy-to-clipboard": "^3.3.2",
     "currency-symbol-map": "^5.1.0",
     "dayjs": "^1.11.6",
-    "embed-api": "https://github.com/wanderwallet/embed-api#3e0be72d8eebf2f0d89d830b71da0be3753564cd",
+    "embed-api": "https://github.com/wanderwallet/embed-api#feat/arc-1198-eddsa-signatures",
```
Danziger: Update before merging.

src/utils/authentication/authentication.types.ts
```diff
@@ -1,21 +1,7 @@
-import type { DbSession, SupabaseAuthError } from "embed-api";
-import type { JwtPayload } from "jwt-decode";
-import type { Provider, Session } from "@supabase/supabase-js";
+import type { SupabaseAuthError } from "embed-api";
+import type { Session } from "@supabase/supabase-js";
 import type { OAuthErrorCode } from "~utils/authentication/authentication.utils";
 
-export type SupabaseJwtSessionData = Omit<DbSession, "id" | "userId">;
```
Danziger, line 6: These types have been moved to the backend and are now available through `embed-api`.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -230,32 +226,6 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
-        createdAt: new Date(),
-        updatedAt: new Date(),
-        userId: "",
-      };
-    }
-
-    const userAgent = navigator.userAgent;
-    const deviceNonce = await getDeviceNonce();
-    // NOTE: We use ipv4 address here as in Vercel backend we get ipv4 address from the request headers.
-    const ip = await getIPAddress().catch(() => session.ip);
```
Danziger, line 266: No need to do this as the IP is not part of the challenge data anymore, so it's ok if the session is out of sync (Supabase session !== DB session).

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1239,68 +1190,22 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
-        }
-
-        if (_event !== "TOKEN_REFRESHED" && (!dbSession.id || dbSession.deviceNonce !== deviceNonce)) {
-          console.log("🔃 Refreshing session...");
-
-          // We wait to leave some time for the trigger to update the session and make sure that, when we refresh, we
-          // get the updated session data:
-          await sleep(2500);
-
-          supabase.auth.refreshSession();
```
Danziger, line 1305: No need to try to force-refresh the session anymore. It's ok if it gets temporarily out of sync.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -1239,68 +1190,22 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
         setEmbeddedContextAuth(({ authStatus }) => ({
           authStatus: authStatus === "unknown" || authStatus === "authError" ? "authLoading" : authStatus,
           authProviderType,
           user,
-          session: dbSession,
+          session,
         }));
       } else {
         // TODO: Duplicated in initEmbeddedWallet()?
+        /*
```
Danziger: TODO: Remove this once tested.

src/utils/embedded/embedded.utils.ts
```diff
@@ -174,31 +171,16 @@ export {
...
-  try {
-    const sessionId = await getSessionId();
     const applicationId = await trpcVanilla.validateApplication.query({
       clientId: EMBEDDED_CLIENT_ID,
       applicationOrigin: EMBEDDED_ANCESTOR_ORIGIN,
       sessionId,
     });
 
-    setApplicationIdHeader(applicationId);
+    // TODO: applicationOrigin: EMBEDDED_ANCESTOR_ORIGIN is not needed. `validateApplication` will return a list of
```
Danziger, line 182: Just a TODO for later, as this function is not used right now.

src/utils/wallets/wallets.service.ts
```diff
@@ -140,27 +140,47 @@ async function fetchFirstAvailableAuthShare(
...
+          generateWalletActivationChallengePromise,
+        ]);
+
+        let shareHash: string = generateShareHashAndEdKeysResult.shareHash;
+        let sharePrivateKey: JWKInterface | Uint8Array = generateShareHashAndEdKeysResult.sharePrivateKey;
+
+        const { activationChallenge } = generateWalletActivationChallengeResult;
+
+        if (activationChallenge.version === "v1") {
+          // If we got a v1 challenge, we still need to migrate this WorkKeyShare to EdDSA. To do that, we first need to
```
Danziger, line 157: Migration logic.

src/utils/wallets/wallets.utils.ts
```diff
@@ -159,75 +161,103 @@ async function generateShareHash(share: string): Promise<string> {
...
-      if (j === share.length) j = 0;
-    }
+  // Salt:
+  //
+  // Using salt adds significantly to the strength of HKDF, but it is not strictly necessary when the input key material
+  // is already sufficiently random, as it should be in our case. However, using a salt won't hurt either.
 
-    return r;
-  };
+  const saltContent = "deviceShare" in params ? `${session.userId}-${session.deviceNonce}` : session.userId;
```
Danziger, line 209: @7i7o See here for salt choices. Note that if `deviceNonce` is deleted or updated, the wallet will have to be recovered, as activation will fail, even if the `deviceShare` is intact.

src/utils/wallets/wallets.utils.ts
```diff
@@ -159,75 +161,103 @@ async function generateShareHash(share: string): Promise<string> {
...
+  deviceShare: string;
+  session: DbSession;
+}
+
+export interface GenerateShareHashAndEdKeysFromRecoveryBackupShareParams {
+  recoveryBackupShare: string;
+  session: DbSession;
+}
+
+export type GenerateShareHashAndEdKeysParams =
```
Danziger, line 174: Used an union here so that we can tell apart what we are going to use this share for based on the param name (`deviceShare` vs `recoveryBackupShare`), as that affects the salt choice.

src/utils/embedded/embedded.provider.tsx
```diff
@@ -96,7 +92,7 @@ const EMBEDDED_CONTEXT_INITIAL_AUTH = {
   authStatus: "unknown",
   authProviderType: null,
   user: null,
-  session: null,
+  session: INITIAL_ANON_SESSION,
```
Danziger: Session is no longer `null`. Instead, we use an empty/anon session object, as `session` is passed to `solveChallenge()`, and `id`, `userAgent` and `deviceNonce` are part of the challenge solution data.

src/utils/embedded/embedded.context.ts
```diff
@@ -0,0 +1,59 @@
...
+} as const satisfies EmbeddedContextState;
+
+export const EMBEDDED_CONTEXT_INITIAL_AUTH = {
+  authStatus: "unknown",
+  authProviderType: null,
+  user: null,
+  session: INITIAL_ANON_SESSION,
+} as const satisfies EmbeddedContextAuth;
+
+export const EmbeddedContext = createContext<EmbeddedContextData>({
```
Danziger, line 26: Moved all these to a separate file to enable hot-reloading. `embedded.provider.tsx` should only export React components.

7i7o (approved):
tACK 👍

package.json
7i7o: ☝️

PR #936, opened 2025-07-15
feat: ARC-1366: Unsupported unpartitioned state banner & screen

Danziger:
Add a banner with the text "Limited browser support" and a separate view with additional details when unpartitioned state is not available / accessible.

The logic to show the "Limited browser support" screen is as follows:

- First visit to a site (not authenticated): The user sees the screen and must click "Continue anyway".
- Subsequent visits to the same site (not authenticated): The user will only see the screen if they marked the "Don't show me this again on this site" checkbox. Note the checkbox has no effect when visiting a new site, they'll see the view again on each new site (due to the lack of unpartitioned state).
- When users click "Learn more" on the "Limited browser support" banner, they'll see the screen again but there will be no "Continue anyway" button or checkbox, just the information.
- The banner is always visible except when in the "Limited browser support" screen.
- If the user is already authenticated, they won't be redirected to the screen (as the fact that they are authenticated implies they've already seen it).

Designs are [here](https://www.figma.com/design/pjzy7KRZrRg1vj6ar8eQi4/Embed-%7C-Wander?node-id=1213-4352&m=dev), but have been adjusted slightly based on the discussion in the comments on those views:

- There's no "unsupported" browsers (Internet Explorer). That is because while IE is not supported, that's not because the lack of unpartitioned state, it would probably be due to the lack of Crypto API. Therefore, it doesn't make sense to mention it here. In any case, we should have a separate screen for totally unsupported browsers, but that would be very rarely used.

- Browsers have been stacked in 2 columns and Opera has been added to "Fully supported.

**Other somewhat related changes:**

- Fix `EnhancedStorage` being instantiated twice for `localStorage` (in `StorageMock`).

- Replace custom event for unpartitioned state status change with callbacks, so that we get the most up to date change as soon as we start listening. Otherwise, the value might have changed after we read its initial value in `useState` but before we start listening for changes, so the event was lost.

- Fix tab text alignment on Safari.

- The back/close/minimize buttons have been updated to use the `Button` component (so they have proper hover and focus effects now).

- The logic in `WalletService.fetchFirstAvailableAuthShare()` hasn been updated to ignore wallets where ` wallet.status !== "ENABLED"`, as that would result in activation errors.

- Update `LocalStorage` and `EnhancedStorage` to cache status and avoid requesting access multiple times in parallel (caching a Promise from the first call to `requestStorageAccess()`).

- Update `EnhancedStorage.requestStorageAccess()` and some other protected/private functions to return the storage status.

- Extract `StorageManager` to its own file.

- The images (screenshots) are cropped a bit on the sides. That's a bug with the `Image` component. I'll be fixing that in this other PR, plus adding support for `source` and replacing the current ones with light/dark versions taken from a different site so that the URLs / page names also look nicer. See https://github.com/wanderwallet/Wander/pull/941.

**Screenshots:**

First visit:

After confirming and subsequent visits:

Wallet home:

After clicking "Learn more":

Access error (Brave):

Access error (Chrome Mobile):

Access error (In-App Android Browser):

Any of the above after 3 failed attempts:

Any of the above after "Continue anyway" and clicking "Learn more" from within the app:

Closes:

- https://communitylabs.atlassian.net/browse/ARC-1366
- https://communitylabs.atlassian.net/browse/ARC-1359
- https://communitylabs.atlassian.net/browse/ARC-1367


src/routes/embedded/support/unpartitioned-state/unpartitioned-state.view.tsx
```diff
@@ -0,0 +1,167 @@
...
+    (unpartitionedStateStatus === "rejected" || unpartitionedStateStatus === "error")
+  ) {
+    if (unpartitionedStateStatus === "rejected") {
+      headerText = "Enable browser storage support";
+    } else {
+      headerText = "Error accessing browser storage";
+    }
+
+    subtitle =
+      "Before you continue, Wander Connect needs access to your browser's storage to enalbe cross-site authentication and wallet synching.";
```
7i7o:
```suggestion
      "Before you continue, Wander Connect needs access to your browser's storage to enable cross-site authentication and wallet synching.";
```

7i7o (approved):
tACK 👍 

Only 1 typo to fix and some routing issues when going back and forth with the limited support screen

Also, on Safari the tabs are not styled properly (but not related to this PR)

Both issues are shown in the video below


https://github.com/user-attachments/assets/08d4bb62-8944-458f-913c-33f1681dce12

src/components/embed/ui/molecules/tabBar/TabBar.module.scss
```diff
@@ -14,6 +14,9 @@
 }
 
 .tabs button {
+  display: flex;
```
Danziger, line 17: Fix the issue with iOS.

src/iframe/storage/plasmo-storage/plasmo-storage.mock.ts
```diff
@@ -47,16 +45,10 @@ export class StorageMock extends PlasmoStorage implements StorageMockInterface {
...
 
-    // This browser doesn't support the Storage Access API
-    // so let's just hope we have access!
-    if (!document.hasStorageAccess) return;
+    this._area = area;
 
-    // TODO: Can this be postponed until authentication to avoid requesting permissions too soon?
-    // unpartitioned sessionStorage cannot be accessed from iframe as it is partitioned by both origin and browser tabs unlike localStorage.
-    if (area === "local") this.storage.requestStorageAccess();
+    this.storage = area === "local" ? LocalStorage.getInstanceWithoutWaiting() : new EnhancedStorage({ area });
```
Danziger, line 51: Before, we ended up with 2 different instances of `LocalStorage`.

src/iframe/storage/storage-manager/storage-manager.ts
```diff
@@ -0,0 +1,308 @@
+import { isComplexStorageItem } from "~iframe/storage/storage-manager/storage-manager.utils";
+
+/**
+ * Storage Manager for handling localStorage and sessionStorage eviction policies
+ */
+export class StorageManager {
```
Danziger, line 6: Just moved this to its own file.

src/iframe/storage/storage-manager/storage-manager.utils.ts
```diff
@@ -0,0 +1,63 @@
+export type StorageItem<T = string> =
```
Danziger, line 1: Just moved these to their own file.

src/iframe/storage/unpartitioned-storage/local-storage.ts
```diff
@@ -24,4 +24,22 @@ export class LocalStorage extends EnhancedStorage {
 
     return this.instance;
   }
+
+  // TODO: This is not the most elegant solution for the use case in `StorageMock` (`plasmo-storage.mock.ts`). Refactor this later:
```
Danziger, line 28: Already commented here. Ideally, all `StorageMock` / `EnhancedStorage` methods simply wait for the unpartitioned state access, but that would now require quite a bit of refactoring, so I just patched it like this.

src/iframe/storage/unpartitioned-storage/unpartitioned-storage.ts
```diff
@@ -525,25 +197,26 @@ export class EnhancedStorage implements Storage {
...
+      this.error = isError(unpartitionedStateStatusOrError)
+        ? unpartitionedStateStatusOrError
+        : new Error("Unexpected unpartitioned state error");
+    } else {
+      this.error = null;
+    }
+
+    log(LOG_GROUP.STORAGE, `Unpartitioned state access for ${this.storageType} = ${unpartitionedStateStatus}`);
+
+    setUnpartitionedStateStatus(unpartitionedStateStatus, this.error);
```
Danziger, line 241: The `UNPARTITIONED_STATE_STATUS_CHANGE_EVENT` event has been replaced with callbacks. `setUnpartitionedStateStatus()` will invoke them. With this change, as soon as you add a callback, it is invoked with the most recent status, fixing a race condition we had before where if we started listening after the status had changed, we won't be notified, so Brave could sometimes show the UI as if it had access to unpartitioned state, when it actually didn't.

src/routes/embedded/account/import-seedphrase/account-import-seedphrase.view.tsx
```diff
@@ -51,8 +51,7 @@ export function AccountImportSeedphraseEmbeddedView() {
       subtitle="Enter your seedphrase to connect your wallet to your account."
       footerElement={<WanderFooter />}
       hasBackButton={true}
-      onBackButtonClick={back}
-      size="auto">
```
Danziger, line 55: Just cleaning up unused props from some components. For `Card`, we always used `auto` (and in the couple of cases where we didn't, we could).

src/routes/embedded/auth/auth/auth.view.tsx
```diff
@@ -187,21 +187,26 @@ export function AuthEmbeddedView() {
...
-      </Row>
+        </Row>
+      ) : (
+        <Button
+          variant="outlined"
+          isFullWidth
+          icon={<GoogleIcon fontSize={24} />}
+          onClick={() => handleAuthenticate("GOOGLE")}
+          isDisabled={areButtonsDisabled}>
+          Continue with Google
```
Danziger, line 207: When only the Google button is shown, it is a bit weird to see the logo with no text, while all the other buttons have logo + text, so I've added it.

src/routes/embedded/support/unpartitioned-state/unpartitioned-state.view.tsx
```diff
@@ -0,0 +1,289 @@
...
+      navigate(EmbeddedPaths.Auth);
+    } catch (error) {
+      // Brave throws a "NotAllowedError" error while trying to request access, even after a user interaction...
+      toast.error("Unexpected error. Please, try again.");
+      setErrorsWhileRequestingAccess((prevErrorsWhileRequestingAccess) => prevErrorsWhileRequestingAccess + 1);
+      setIsRequestingPermission(false);
+    }
+  }, [navigate]);
+
+  useInterval(
```
Danziger, line 96: With this, the user doesn't need to click the button. Still, the button is there just in case and to trigger the "Continue anyway" flow if for whatever reason we cannot get access.

src/utils/embedded/utils/routes/embedded-routes.utils.ts
```diff
@@ -9,7 +11,11 @@ const routeTypeByLocationPrefix: Record<string, RouteType> = {
   "auth-request": "auth-request",
 };
 
-export function locationToRouteType(path: WanderRoutePath): RouteType {
+export function locationToRouteType(path: WanderRoutePath, authStatus: AuthStatus): RouteType {
+  if (path === EmbeddedPaths.SupportUnpartitionedStateMissing) {
+    return authStatus === "unlocked" ? "default" : "auth";
```
Danziger, line 16: This view can be shown before auth and after auth, so we change the layout to match the rest of the views.

src/utils/wallets/wallets.service.ts
```diff
@@ -134,7 +134,7 @@ async function fetchFirstAvailableAuthShare(
       const { id: walletId, address: walletAddress, deviceShare } = wallet;
 
       try {
-        if (deviceShare === null) {
+        if (deviceShare === null || wallet.status !== "ENABLED") {
```
Danziger, line 131: Fix for the issue reported by Odysee last Friday.

Danziger:
In Brave, I've tried the following:

- Click "Try again" 3 times without giving permissions.
- Click "Continue anyway".
- Log out.
- Reload.
- Give access.
- Reload.
- Log in.

I thought by doing that, the key share would have been stored in the partitioned state first, and after giving access it would try to get it from the unpartitioned one, and not find it. It turns out that's not the case and it can find it just fine.

I also tried the other way around: Give access first, then remove it while still using the app. That doesn't seem to cause any issue either.

Note our current implementation doesn't subscribe to permission changes in most cases, so it will not get notified when they are removed. But again, that doesn't seem to be an issue, so we don't need to do anything.

PR #939, opened 2025-07-17
Fix: Show error icon and tooltip when token fails to return balance

pawanpaudel93:
## Summary

This PR fixes the following:
1. Shows the error icon and tooltip when a token process fails to return a balance. Previously, failures showed 0 with no context—this fix brings back the proper UI feedback for better user clarity. Closes [ARC-1408](https://communitylabs.atlassian.net/browse/ARC-1408)
2. Fixes height shrinking issue with the token selector slider in send flow
3. Show token ticker instead of token name in header of `Menu -> Tokens -> Select a token` as there can be very long names for tokens.

## How to test

1. Turn off Wi-Fi and open the extension from both this branch build and the production build. Check if this branch shows error indicators when balances fail to load.
2. Check if any token balance fails to load and verify that an error tooltip with the message `Token balance error` appears.
3. You can also manually trigger an error by throwing inside `fetchTokenBalance` here and see the `Token balance error` if none of the token balance fetching failed: https://github.com/wanderwallet/Wander/blob/de55b1eb83a312c3600aff65f93e5e0c2cc4dc5c/src/tokens/aoTokens/ao.ts#L474
4. Make sure the token selector slider in send flow is looking good (not shrinked) when it has lot of tokens.

[ARC-1408]: https://communitylabs.atlassian.net/browse/ARC-1408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,7 +168,7 @@
     "human-crypto-keys/crypto-key-composer/node-forge": "^1.3.1",
     "axios": "^1.7.2",
     "rollup": "4.43.0",
-    "string-width": "4.2.3"
+    "@inquirer/core/wrap-ansi/string-width": "4.2.3"
   },
   "lint-staged": {
     "*.{js,ts,tsx,html,css,scss,md,json}": "prettier --write"
--- a/package.json
+++ b/package.json
@@ -167,7 +167,8 @@
     "human-crypto-keys/node-forge": "^1.3.1",
     "human-crypto-keys/crypto-key-composer/node-forge": "^1.3.1",
     "axios": "^1.7.2",
-    "rollup": "4.43.0"
+    "rollup": "4.43.0",
+    "string-width": "4.2.3"
   },
   "lint-staged": {
     "*.{js,ts,tsx,html,css,scss,md,json}": "prettier --write"
--- a/src/tokens/hooks.ts
+++ b/src/tokens/hooks.ts
@@ -25,8 +25,8 @@ import { retryWithDelay } from "~utils/promises/retry";
 
 export const defaultOptions = {
   refetchInterval: 300_000,
-  // staleTime: 300_000,
-  // gcTime: 300_000,
+  staleTime: 300_000,
+  gcTime: 300_000,
   retry: 3,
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -167,7 +167,8 @@
     "human-crypto-keys/node-forge": "^1.3.1",
     "human-crypto-keys/crypto-key-composer/node-forge": "^1.3.1",
     "axios": "^1.7.2",
-    "rollup": "4.43.0"
+    "rollup": "4.43.0",
+    "@inquirer/core/wrap-ansi/string-width": "4.2.3"
```
pawanpaudel93, line 171: Was having error `[ERR_REQUIRE_ESM]: require() of ES Module` so I put a version that doesn't throw that error. To see the error you can try to run the dev build.

src/tokens/hooks.ts
```diff
@@ -84,13 +85,19 @@ export function useTokenBalance(token: TokenInfo, address: string, refresh?: boo
     queryKey: ["tokenBalance", token.processId, address],
     queryFn: async () => {
       try {
-        const balance = await fetchTokenBalance(token, address, refresh);
+        const balance = await retryWithDelay(
```
pawanpaudel93, line 88: Disabled tanstak retry as it was acting weird and not working as expected. So, used the custom retry inside queryFn and now if there is any error from fetchTokenBalance regarding network or balance fetch, it would be correctly shown with the error tooltips for tokens.

src/utils/tanstack.ts
```diff
@@ -1,3 +1,12 @@
 import { QueryClient } from "@tanstack/react-query";
 
-export const queryClient = new QueryClient();
+export const queryClient = new QueryClient({
```
pawanpaudel93, line 3: Now, we use this tanstack query client instance from here everywhere.

package.json
7i7o: As discussed, we'll fix this next week by updating dev, staging and prod envs to latest Node LTS

7i7o (approved):
tACK 👍

PR #941, opened 2025-07-17
feat: Better image component with sources

Danziger:
`Image` component fixes / improvements:

- Use inline elements (`picture`) rather than block (`div`), so that we don't get HTML errors when placing `Image` in a place where a regular `img` or `picture` tag could be placed (like inside a `p`).
- Fix incorrectly cropping of images on the sides.
- Add light/dark image support.
- Limit `borderRadius` to `"none" | "rounded" | "circular"`.
- Add `fullWidth` prop for variable-size images that maintain their aspect-ratio.
- Add a fake delay when loading images in development (so that we can see the placeholder).
- Make `alt` required.
- Make `placeholderURL` optional (images can have no visible placeholder but still "reserve" their space).

This PR also introduces a new `TokenLogo` component that abstract loading and displaying an AO token or AR logo (in his latter case, changing depending on theme). Additionally:

- All previous instances where we were displaying one such logo have been updated.
- Both `logo_light.png` and `logo_dark.png` have been deleted and replaced by `TokenLogo` too.
- Instances where we had to load either the light or dark AR logo version have been updated to use the SVGs instead of the old (now deleted) PNGs.

Some other fixes:

- `BatchSignDataItemAuthRequestView` and `SignDataItemDetails` were using `class Token` (from `ao-tokens`) instead of `fetchTokenByProcessId()`, causing many request to load token info to fail. That has been updated and a new issue has been created to replace the remaining usages of `Token` and also DRY the fallback logic: https://communitylabs.atlassian.net/browse/ARC-1530, https://communitylabs.atlassian.net/browse/ARC-1531

- `SignDataItemDetails` has been exported using a named export and its file renamed from `index.ts` to `SignDataItemDetails.tsx`, just like with its Connect counterpart.

- Some inconsistencies have been detected between `signDataItem` and `batchSignDataItem`'s screens between BE and Connect. The issue with the token logos not loading was due to a missing `dataItem.target` property rather than due to an issue in the implementation, though. Some TODOs have been added to the code and a new issue has been created to address that later: https://communitylabs.atlassian.net/browse/ARC-1530

**Screenshots:**

**Brave:**

**Chrome Mobile:**

**In App Browser:**


Danziger:
Waiting for https://github.com/wanderwallet/Wander/pull/936 to be merged in order to try the new `srcDark` in there.

src/components/common/Image/Image.tsx
```diff
@@ -0,0 +1,136 @@
...
+    if (fakeDelay) await sleep(fakeDelay + Math.random() * fakeDelay * 2);
+
+    setIsLoaded(true);
+  };
+
+  const [devPointer, setDevPointer] = useState<[number, number] | null>(null);
+
+  const pointer = devPointer || pointerProp;
+
+  const movePointer =
```
Danziger, line 86: In development, clicking on an image will show a "click indicator" on it and log its position in the console for you to copy it and add it to the code.

src/components/common/Image/Image.tsx
```diff
@@ -0,0 +1,136 @@
...
+          console.log("pointer =", nextPointer);
+
+          setDevPointer(nextPointer);
+        }
+      : undefined;
+
+  return (
+    <picture className={pictureClassName} style={pictureStyle}>
+      {srcDark ? (
+        <source srcSet={displayTheme === "light" ? src : srcDark} media="(prefers-color-scheme: dark)" />
```
Danziger, line 105: This can easily be extended later to support more modern formats and different sizes/resolutions. Not doing any of that now to avoid making this PR longer and because we'd probably want to look into automating that if possible.

src/components/dashboard/list/AppListItem.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { Image } from "~components/common/Image/Image";
 
 export default function AppListItem({ name, url, icon, active, ...props }: Props & HTMLProps<HTMLDivElement>) {
   return (
     <ListItem
       title={name}
       subtitle={url}
       active={active}
-      icon={icon && <Image height={40} width={40} src={icon} />}
+      icon={icon && <Image height={40} width={40} src={icon} placeholderURL="none" alt={`${name} logo`} />}
```
Danziger, line 12: No placeholder because these images can have any shape. We could consider extending the placeholder property of the `Image` component to support a more subtle placeholder for these cases (small semitransparent circle, for example).

src/components/dashboard/list/TokenListItem.tsx
```diff
@@ -1,20 +1,15 @@
...
-import arLogoDark from "url:/assets/ar/logo_dark.png";
-import { getUserAvatar } from "~lib/avatar";
 import { useLocation } from "~wallets/router/router.utils";
-import CommonImage from "~components/common/Image";
 import { AR_PROCESS_ID } from "~tokens/aoTokens/ao";
-import { useAsyncEffect } from "~utils/react/useAsyncEffect";
+import { TokenLogo } from "~components/popup/TokenLogo";
+
+import aoLogo from "url:/assets/ecosystem/ao-logo.svg";
+import arLogo from "url:/assets/ecosystem/ar-logo.svg";
```
Danziger, line 12: Always shown in the same color, so importing it directly rather than using `TokenLogo`.

src/components/popup/home/WalletActions.tsx
```diff
@@ -1,9 +1,10 @@
 import { ActionButtons, type ButtonConfig } from "../ActionButtons";
 import { QrCode02 } from "@untitled-ui/icons-react";
 import browser from "webextension-polyfill";
-import arLogoDark from "url:/assets/ar/logo_dark.png";
 import { ReceiveIcon } from "~components/icons/ReceiveIcon";
 
+import arLogoDark from "url:/assets/ar/ar-logo-dark.svg";
```
Danziger, line 6: Replaced with SVG version.

src/components/popup/Token.tsx
```diff
@@ -248,50 +228,6 @@ export const LogoAndDetails = styled.div`
...
-  position: relative;
-  flex-shrink: 0;
-  border-radius: 29px;
-  object-fit: cover;
-  overflow: hidden;
-  z-index: 1;
-  width: ${(props) => (props.width ? `${props.width}px` : "40px")};
-  height: ${(props) => (props.height ? `${props.height}px` : "40px")};
-
-  &::before {
```
Danziger, line 275: The background fill is no longer needed because the AR logo properly changes to dark or light depending on the theme.

For other tokens, I'd argue they should not be expecting their provided image to be filled in with background colors, so it should be good too.

src/components/popup/TokenLogo.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { getUserAvatar } from "~lib/avatar";
+import type { Token } from "~tokens/token";
+import { AR_LOGO, AR_PROCESS_ID, type TokenInfo } from "~tokens/aoTokens/ao";
+import { FULL_HISTORY, useGateway } from "~gateways/wayfinder";
+import { concatGatewayURL } from "~gateways/utils";
+
+import arLogoLight from "url:/assets/ar/ar-logo-light.svg";
+import arLogoDark from "url:/assets/ar/ar-logo-dark.svg";
+
+function isAr(token: string | Partial<Token>) {
```
Danziger, line 13: Before, we were showing different versions of the AR logo in different places:

- PNG versions from the repo.
- SVG versions from the repo.
- Image from Arweave (`AR_LOGO`).

Now, this component handles all those different values referring to the same thing in a consistent way, showing the repo light and dark SVGs.

src/routes/embedded/auth/import-seedphrase/auth-import-seedphrase.view.tsx
```diff
@@ -66,12 +66,6 @@ export function AuthImportSeedphraseEmbeddedView() {
         return;
       }
 
-      if (wallets.length > 0) {
```
Danziger, line 69: This was preventing importing a seedphrase as part of the recovery flow (to generate a new wallet).

src/routes/popup/transaction/[id].tsx
```diff
@@ -246,41 +248,45 @@ export function TransactionView({ params: { id, gateway: gw, message } }: Transa
...
+                setTokenInfo(tokenInfo);
+
                 data.transaction.quantity = {
                   ar: amount.toFixed(),
                   winston: "",
                 };
                 data.transaction.recipient = aoRecipient;
               } else {
-                setLogo(arweaveLogo);
+                // TODO: Should this case ever happen?
```
Comment (Danziger) [inline src/routes/popup/transaction/[id].tsx:267]: cc: @7i7o

src/tokens/token.ts
```diff
@@ -20,41 +15,6 @@ export interface Token {
...
-/**
- * Load token logo from Viewblock. If no logo
- * is returned, return the original logo from
- * the contract.
- *
- * @param id Contract ID of the token
- * @param defaultLogo Default logo tx ID in the contract state
- * @param theme UI theme to match the logo with
- */
-export async function loadTokenLogo(id: string, defaultLogo?: string, theme?: DisplayTheme) {
```
Danziger, line 32: This function was only used in 1 place and you can see the TSDoc doesn't match what it does, so I guess it's an old version that was left behind. Everywhere else, `getUserAvatar` was used.

src/components/embed/auth/SignDataItemDetails.tsx
```diff
@@ -10,70 +10,67 @@ import prettyBytes from "pretty-bytes";
...
-import arLogoLight from "url:/assets/ar/logo_light.png";
 import { Box, ChevronRight, Spacer, Text, Row } from "../ui";
 import TransactionTag from "./TransactionTag";
+import { useAsyncEffect } from "~utils/react/useAsyncEffect";
+import { TokenLogo } from "~components/popup/TokenLogo";
+import type { RawDataItem } from "~api/modules/sign_data_item/types";
 
-export default function SignDataItemDetails({ params }) {
+export interface SignDataItemDetailsProps {
+  dataItem: RawDataItem;
```
Danziger, line 20: As I was reviewing all this, I've done some small improvements here and there. Like adding this missing type.

src/components/embed/ui/molecules/card/auth-request-card/AuthRequestCard.tsx
```diff
@@ -72,7 +72,7 @@ export function AuthRequestCard({
       {...cardProps}
       hasBackButton={!!onBackButtonClick}
       onBackButtonClick={onBackButtonClick}
-      customIcon={<XClose fontSize={24} color={"#666666"} />}
```
Danziger, line 75: This was incorrectly placed do to a `margin-right`.

src/components/SignDataItemsDetails.tsx
```diff
@@ -9,81 +9,75 @@ import {
...
-  const recipient = params?.tags?.find((tag) => tag.name === "Recipient")?.value || "";
-  const quantity = params?.tags?.find((tag) => tag.name === "Quantity")?.value || "0";
-  const transfer = params?.tags?.some((tag) => tag.name === "Action" && tag.value === "Transfer");
+  const recipient = dataItem?.tags?.find((tag) => tag.name === "Recipient")?.value || "";
+  const quantity = dataItem?.tags?.find((tag) => tag.name === "Quantity")?.value || "0";
+  const transfer = dataItem?.tags?.some((tag) => tag.name === "Action" && tag.value === "Transfer");
+  const tokenName = tokenInfo?.Name;
+
+  useAsyncEffect(async () => {
+    // TODO: See if dataItem with no `target` property but a Target tag is valid, and update this code if needed.
```
Danziger, line 42: @7i7o See here about what we discussed. If this property is added, everything works. I'm just wondering if it should also work without it as long as the `Target` tag is added. We'll probably have to test with an actual transaction and see what happens.

src/routes/auth/batchSignDataItem.tsx
```diff
@@ -81,6 +82,9 @@ export function BatchSignDataItemAuthRequestView() {
                 name = item.target;
               }
             }
+
+            // TODO: Add the token logo or a "data" icon next to each item:
```
Danziger, line 86: Just added a TODO here for now, and created a new issue on Jira to first check with design how each of these views should look in BE vs Connect. Then we can adjust accordingly as right now for `batchSignDataItem` we are using an adjusted version of what's on BE, rather than a design similar to the `signDataItem` one from Connect.

src/routes/auth/signDataItem.tsx
```diff
@@ -130,53 +124,51 @@ export function SignDataItemAuthRequestView() {
...
-      let tokenInfo: TokenInfo;
+  useAsyncEffect(async () => {
+    if (!process || !transfer) return;
+
+    let tokenInfo: TokenInfo;
+
+    try {
+      setLoading(true);
+
+      const token = await Token(data.target);
```
Danziger, line 142: I've created a new issue to replace all usages: https://communitylabs.atlassian.net/browse/ARC-1529

The one for `batchSignDataItem` I've already done it, as we do multiple requests in parallel there so it would always fail.

src/routes/auth/connect.tsx
```diff
@@ -455,12 +455,13 @@ const ConfirmPage = ({
 }) => (
   <ConnectPageContent>
     <Section showPaddingVertical={false} style={{ display: "flex", flexDirection: "column", gap: "24px" }}>
-      <div style={{ textAlign: "center" }}>
+      <div style={{ display: "flex", flexDirection: "column", gap: "8px" }}>
         <PrimaryText fontSize={20} fontWeight={600}>
           {browser.i18n.getMessage("confirm_permissions", [appInfo.name || url])}
         </PrimaryText>
-        <SecondaryText>{url}</SecondaryText>
+        {appInfo.name !== url ? <SecondaryText>{url}</SecondaryText> : null}
```
Danziger, line 462: Otherwise we just show the same thing twice.

src/components/popup/TokenLogo.tsx
```diff
@@ -0,0 +1,179 @@
...
+    ? token.toUpperCase() === AR_PROCESS_ID || token === AR_LOGO || token.endsWith(AR_LOGO)
+    : token.id === AR_PROCESS_ID || token.ticker === "AR" || token.name === "Arweave" || token.defaultLogo === AR_LOGO;
+}
+
+function isURI(token: string | Partial<Token>): token is string {
+  return typeof token === "string" && (token.startsWith("chrome-extension://") || token.startsWith("https://"));
+}
+
+function getTokenFallbackImage(token: string | Partial<Token>, name: string) {
+  const fallbackSymbol = (typeof token === "object" && token.ticker) || (name.includes(" ") ? "" : name) || "?";
```
pawanpaudel93, line 24: `name.includes` is failing for `SignDataItemAuthRequestView` popup when sending AO tokens loading the token image. Also, there was token image on a squircle in that view and the squircle is missing now.
image

Danziger: Fixed.

Regarding the squircle, I've removed that one because it's the only place where I saw a squircle with a token logo. All other token logos are circular, so it's probably better to keep it consistent.

7i7o (approved):
tACK 👍

PR #945, opened 2025-07-21
feat: Enforce one wallet per user.

Danziger:
The logic to do this has been moved to the backend. See https://github.com/wanderwallet/embed-api/pull/49/files.

Merge backend first. Then immediately update `embed-api` commit hash here and merge (as the `updateInfo` function has been removed from the backend).


7i7o (approved):
utACK 👍

PR #948, opened 2025-07-22
feat: use ar.io sdk for ArNS operations

7i7o:
Adds ar.io SDK and replaces custom lookup code and types for ArNS with ones from SDK.

Replaces and closes #876 authored by @kunstmusik

Some comments are in the original PR


pawanpaudel93:
@7i7o I think we should run `yarn` again to install dependencies as I can see changes to `yarn.lock` after running `yarn`.

src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
```diff
@@ -18,7 +18,7 @@ export async function handleGatewayUpdateAlarm(alarm?: Alarms.Alarm) {
 
   try {
     // fetch cache
-    const ArIO = new AOProcess({ processId: AO_ARNS_PROCESS });
+    const ArIO = new AOProcess({ processId: ARIO_MAINNET_PROCESS_ID });
```
pawanpaudel93: `AOProcess` should be imported from `@ar.io/sdk` and I think we can remove `AOProcess` implementation from `import { AOProcess } from "~lib/ao"`

src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
```diff
@@ -18,7 +18,7 @@ export async function handleGatewayUpdateAlarm(alarm?: Alarms.Alarm) {
...
 
     const gatewaysResult = await ArIO.read<PaginatedResult<GatewayAddressRegistryItemData>>({
```
pawanpaudel93: I think we can use https://github.com/ar-io/ar-io-sdk?tab=readme-ov-file#getgateways-cursor-limit-sortby-sortorder- instead of using AOProcess to query the gateways.

src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
7i7o: Addressed in [last commit](https://github.com/wanderwallet/Wander/pull/948/commits/df0cbf189a2aebe6d12e66ee064415887e6068ee)

src/api/background/handlers/alarms/gateway-update/gateway-update-alarm.handler.ts
7i7o: Addressed in [last commit](https://github.com/wanderwallet/Wander/pull/948/commits/df0cbf189a2aebe6d12e66ee064415887e6068ee)

src/lib/wayfinder.ts
```diff
@@ -65,7 +65,7 @@ async function pingUpdater(
         newData[index].health = { status: "pending" };
         onUpdate(newData);
 
-        const healthJson = await fetchResult.json();
+        // const healthJson = await fetchResult.json();
```
pawanpaudel93: Aren't we checking the health now or it is not needed now ?

pawanpaudel93 (approved):

src/lib/wayfinder.ts
7i7o: No, I commented that out while testing, will revert now

PR #965, opened 2025-07-31
fix: Add patch script for math-intrinsics

pawanpaudel93:
## Summary

This is a temporary fix for development build error until Plasmo upgrades to the latest Parcel or the project moves to a monorepo setup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "homepage": "https://www.wander.app",
   "scripts": {
     "dev:chrome": "yarn patch:math-intrinsics && cross-env PARCEL_WORKER_BACKEND=process plasmo dev --verbose",
-    "build:chrome": "cross-env PARCEL_WORKER_BACKEND=process plasmo build --no-hoist",
+    "build:chrome": "yarn patch:math-intrinsics && cross-env PARCEL_WORKER_BACKEND=process plasmo build --no-hoist",
     "dev:firefox": "plasmo dev --target=firefox-mv2",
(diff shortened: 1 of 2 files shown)


patches/patch-math-intrinsics.js
```diff
@@ -0,0 +1,24 @@
+const fs = require("fs");
+const path = require("path");
+
+// Skip if in production mode
+if (process.env.NODE_ENV === "production") {
+  console.log("🚫 Skipping patch: NODE_ENV is production");
+  process.exit(0);
+}
```
7i7o: If we don't patch it, the build fails so we won't be able to build prod

package.json
```diff
@@ -7,7 +7,7 @@
   "packageManager": "yarn@1.22.18",
   "homepage": "https://www.wander.app",
   "scripts": {
-    "dev:chrome": "cross-env PARCEL_WORKER_BACKEND=process plasmo dev --verbose",
+    "dev:chrome": "yarn patch:math-intrinsics && cross-env PARCEL_WORKER_BACKEND=process plasmo dev --verbose",
```
7i7o, line 10: I think the patch should be for every build, not just `dev`
It fails to build for prod if we don't patch

7i7o (changes requested):

package.json
pawanpaudel93: @7i7o Updated for prod build as well.

7i7o (approved):
tACK 👍

PR #970, opened 2025-08-01
Fix transferable pi + update Astro announcement copy

7i7o:
This PR updates the copy for the Astro announcement and removes PI from the non-transferable token list


pawanpaudel93 (approved):
tACK 🚀

I think we can also remove `PI_token_send_popup` from the messages as well.

7i7o:
> tACK 🚀
> 
> I think we can also remove `PI_token_send_popup` from the messages as well.

Removed 🫡

PR #975, opened 2025-08-06
feat: add analytics for earn-related events and pages

pawanpaudel93:
## Summary

This PR adds the events and page analytics for earn screens.


src/routes/popup/earn/manage.tsx
```diff
@@ -126,6 +127,17 @@ export function ManageEarningsView() {
...
+      const changedTokens = tokens.filter((t) => t.flpId in changedDelegations);
+      const tokenIds = changedTokens.map((t) => t.processId);
+      const flpIds = changedTokens.map((t) => t.flpId);
+
+      await trackEvent(EventType.ALLOCATION_UPDATE, {
+        tokenNames: changedTokens.map((t) => t.Name).join(","),
+        tokenProcessIds: tokenIds.join(","),
+        tokenPercentages: flpIds.map((id) => `${changedDelegations[id]}%`).join(","),
+      });
+
```
7i7o: I think we should track analytics in its own try/catch or maybe at the end of this one so the user gets the "updated delegations" toast even if for whatever reason the analytics fails

7i7o (changes requested):
Just a minor request

src/routes/popup/earn/manage.tsx
pawanpaudel93: @7i7o Addressed here https://github.com/wanderwallet/Wander/pull/975/commits/a102dc4a3e6bd49d5b53c72d9641fc431f59dcdc

pawanpaudel93: @7i7o should i remove await for trackEvent so it doesnot block the flow until it tracks the event ?

7i7o: That's an option, but then you wouldn't know if the event tracking throws any errors

pawanpaudel93: Yes, we won't know if it throw any error but its working right now.

7i7o (approved):
tACK 👍

PR #983, opened 2025-08-08
feat: Handle third party cookies off

Danziger:
The logic in `EnhancedStorage` has been updated to avoid registering duplicate click and permission change listeners.

Additionally, when the storage permissions change:

- Authenticated users will be de-authenticated and the app reloaded.
- Non-authenticated users with a Supabase auth token stored in the newly accessible `localStorage` are re-authenticated, by reloading the app.

The flow of the `UnpartitionedStateMissingEmbeddedView` view has been updated to:

- Only show the "Check access" button first.
- After one attempt, the first information block appears.
- For some browsers, a link that says "Not there? Try this instead." appears at the end of the information block.
- Clicking that link expands a second option to give the requested permissions to the app.
- At the end of the second block, we briefly mention that could also be achieved from the browser settings (or Brave shields).

**Check access only:**

**First info block:**

Appears after clicking "Check access" for the first time:

Expands after clicking "Not there? Try this instead." for the first time:

**Second info block:**

 

Note: The screenshots and texts change from browser to browser. In Brave, for example, the last paragraph says "Alternatively, turning Shields down should also work.". Also, Chrome's version is used as fallback, so other browsers will also see that (e.g. Opera).

Also, for browsers that support unpartitioned state but where we don't have access, users now will have to go through this flow to skip giving access:

- After clicking "Check access" for the first time, it updates to "Check access again".
- Clicking it again makes a second option appear: "Continue without".
- Clicking that takes users here, where they can confirm they want to continue without by clicking the same button again, or go back to the previous view to try to get access:

Once the user is already authenticated, this view is accessible from the warning at the top of each screen:

And changes to one of these 2 options:

If users ignore the warning and change the permissions from the browser, they are signed out automatically, potentially losing access to their wallet if it was not backed up. However, settings the permissions back to what they were and reloading again will probably fix it.

**Further refinements:**

Most browsers will include the unpartitioned state access in the third-party cookies permission, thus the current copy and screenshots. However, in some cases, the browser will prompt the user for "embedded content" (preferred):

Note: This looks better on Chrome. On Edge, is a bit broken.

This cases is accounted for in the "Check access only" option above. When the users click the button, the browser will automatically open the permission popup, and if the user accepts, there's no need for any additional instructions.

In some other cases, the browser will show an "embedded content" icon in the navigation bar, where they user can give the app more granular permissions:

Note: Again, in Edge this is slightly broken (red toggle).

We most likely want this to be the first instructions option we show, followed by the third-party cookies icon on the right, followed by the thid-party cookies icon on the left (site info popup), followed by any fallback indications, like going into settings or Brave shields.

In order to better accommodate all this, we probably want to add an accordion or tabs. After discussing this with design, we'll revisit this.


src/utils/embedded/embedded.provider.tsx
```diff
@@ -186,6 +175,54 @@ export function EmbeddedProvider({ children }: EmbeddedProviderProps) {
...
+  // Unpartitioned state:
+
+  useEffect(() => {
+    async function handleUnpartitionedStateStatusChange({
+      unpartitionedStateStatus,
+      prevUnpartitionedStateStatus,
+    }: UnpartitionedStateStatusChangeData) {
+      setUnpartitionedStateStatus(unpartitionedStateStatus);
+
+      // We want to prevent the app from using data from the partitioned state after the permissions has been granted, or from the unpartitioned state once the
```
Danziger, line 187: See here.

src/iframe/storage/unpartitioned-storage/unpartitioned-storage.ts
```diff
@@ -182,31 +162,66 @@ export class EnhancedStorage implements Storage {
...
-        await this.requestStorageAccess();
+        this.permissionStatusPromise = navigator.permissions.query({
+          name: "storage-access" as PermissionName,
+        });
+
+        permissionStatus = await this.permissionStatusPromise;
+        permissionState = permissionStatus.state;
 
-        log(LOG_GROUP.STORAGE, "Storage access granted after user interaction");
+        permissionStatus.addEventListener("change", () => {
```
Danziger: Before, we would only listen for permission changes if `document.hasStorageAccess()` returned `false`. Now we always do, because the permissions can change in any direction.

src/iframe/storage/unpartitioned-storage/unpartitioned-storage.ts
```diff
@@ -182,31 +162,66 @@ export class EnhancedStorage implements Storage {
...
+  }
+
+  /**
+   * Set up a handler to request storage access on user interaction
+   * This is required by the Storage Access API for security
+   */
+  private setupUserInteractionHandler(): void {
+    if (this.isWaitingForUserAction) return;
+
+    this.isWaitingForUserAction = true;
```
Danziger, line 216: We were were adding many listeners here. Only one should be added. Same for the permissions change event above.

src/routes/embedded/support/unpartitioned-state/unpartitioned-state.view.tsx
```diff
@@ -83,26 +97,36 @@ export function UnpartitionedStateMissingEmbeddedView() {
...
 
       navigate(EmbeddedPaths.Auth);
     } catch (error) {
       // Brave throws a "NotAllowedError" error while trying to request access, even after a user interaction...
-      toast.error("Unexpected error. Please, try again.");
+
+      toast.error(
+        error instanceof Error &&
+          (error.name === "NotAllowedError" || error.message === COULD_NOT_ACCESS_UNPARTITIONED_STATE_ERR_MESSAGE)
+          ? "Could not get access. Did you enable it?"
```
Danziger, line 132: Friendlier and probably more accurate message.

src/routes/embedded/support/unpartitioned-state/unpartitioned-state.view.tsx
```diff
@@ -137,14 +161,13 @@ export function UnpartitionedStateMissingEmbeddedView() {
   }, [unpartitionedStateStatus]);
 
   const needsConfirmation = !unpartitionedStateConfirmed && authStatus === "noAuth";
-  const requestAccessButtonText =
-    unpartitionedStateStatus === "rejected" || errorsWhileRequestingAccess === 0 ? "Request access" : "Try again";
+  const requestAccessButtonText = errorsWhileRequestingAccess === 0 ? "Check access" : "Check access again";
+  const accessTo = browserInfo.isBrave ? "embedded content" : "third-party cookies";
```
Danziger, line 171: Replaced the previous wording with more accurate one, as most users that end up in this view are aware of these privacy-related terms. Also, it's confusing to mention made-up terms like "cross site wallet syncing" and then show screenshots below that mention none of that.

src/routes/embedded/support/unpartitioned-state/unpartitioned-state.view.tsx
```diff
@@ -137,14 +161,13 @@ export function UnpartitionedStateMissingEmbeddedView() {
...
 
   let headerText = "Limited browser support";
-  let subtitle =
-    "Your browser doesn't support cross-site authentication and wallet syncing. You'll need to manually import your wallet on each new site.";
-
-  // Logos downloaded from https://github.com/alrra/browser-logos
+  let subtitle = couldProbablyGetAccess
+    ? `Wander could not get access to ${accessTo}. You'll need to manually import your wallet on each new site.`
+    : "Your browser doesn't support cross-site wallet syncing. You'll need to manually import your wallet on each new site.";
```
Danziger, line 176: Here "cross-site wallet syncing" has been left in place, because this option is shown to browsers that do not support unpartitioned state, so no screenshots or additional information is show in this case.

pawanpaudel93:
On macOS Chrome 139.0.7258.68, even with third-party cookies allowed, the “Enable third-party cookies” screen still appears.

third-party-cookies

Danziger:
> On macOS Chrome 139.0.7258.68, even with third-party cookies allowed, the “Enable third-party cookies” screen still appears.
> 
> third-party-cookies

I'll be debugging this later (after the monorepo work is completed) with you or @7i7o, as I don't have a macOS device to test.

7i7o:
> > On macOS Chrome 139.0.7258.68, even with third-party cookies allowed, the “Enable third-party cookies” screen still appears.
> > third-party-cookies
> 
> I'll be debugging this later (after the monorepo work is completed) with you or @7i7o, as I don't have a macOS device to test.

Same comment as Pawan, now my default Chrome is asking me for 3rd party cookies access even when I have everything turned on (cookies & embed content)

7i7o (approved):
ACK 👍 

Only comment is now we need to solve the false positive for 3rd party cookies blocked on MacOS Chrome

PR #984, opened 2025-08-11
fix: tier WNDR balance overflow

pawanpaudel93:
## Summary

This PR fixes the WNDR balance overflow issue in`Your tier` page.

| Before | After |
| --- | --- |
| Before | After |

### DIFF
--- a/src/routes/popup/tier.tsx
+++ b/src/routes/popup/tier.tsx
@@ -31,6 +31,7 @@ import { Link } from "~components/common/Link";
 
 const stars = defaultStars.toSpliced(1, 1);
 const wanderTokenInfo = defaultTokens[3];
+const ALPHABET_REGEX = /[a-z]/i;
 
 export function TierView() {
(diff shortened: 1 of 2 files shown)


src/routes/popup/tier.tsx
```diff
@@ -39,12 +40,23 @@ export function TierView() {
   const { data: activeTier } = useActiveTier();
 
   const formattedBalance = useMemo(() => {
+    if (!activeTier) return { displayBalance: "0", tooltipBalance: "0", showTooltip: false };
     const fractionedBalance = balanceToFractioned(String(activeTier?.balance ?? 0), {
       decimals: wanderTokenInfo.Denomination,
     });
     return formatBalance(fractionedBalance);
```
pawanpaudel93, line 46: You can play around with `fractionedBalance` value to check if the balance is not overflowing. It usually happens with very low balances like less than 0.

7i7o (approved):
tACK 👍

PR #985, opened 2025-08-11
fix: Fix theme providers mess

Danziger:


src/components/popup/TokenLogo.tsx
```diff
@@ -7,13 +7,17 @@ import { AR_LOGO, AR_PROCESS_ID, type TokenInfo } from "~tokens/aoTokens/ao";
...
 
+import tokenPlaceholder from "url:/assets/images/tokens/loading-token.svg?no-inline";
 import arLogoLight from "url:/assets/ar/ar-logo-light.svg";
 import arLogoDark from "url:/assets/ar/ar-logo-dark.svg";
 
 function isAr(token: string | Partial<Token>) {
   return typeof token === "string"
     ? token.toUpperCase() === AR_PROCESS_ID || token === AR_LOGO || token.endsWith(AR_LOGO)
-    : token.id === AR_PROCESS_ID || token.ticker === "AR" || token.name === "Arweave" || token.defaultLogo === AR_LOGO;
+    : token.id === AR_PROCESS_ID || token.defaultLogo === AR_LOGO;
```
Danziger, line 17: Removed 2 options here to avoid the AR logo showing up in "fake" tokens with the same ticker or name.

src/components/popup/TokenLogo.tsx
```diff
@@ -7,13 +7,17 @@ import { AR_LOGO, AR_PROCESS_ID, type TokenInfo } from "~tokens/aoTokens/ao";
 import { FULL_HISTORY, useGateway } from "~gateways/wayfinder";
 import { concatGatewayURL } from "~gateways/utils";
 
+import tokenPlaceholder from "url:/assets/images/tokens/loading-token.svg?no-inline";
```
Danziger, line 10: Custom placeholder for tokens. Updated the color schema to match the `Skeleton` component used in the tokens list.

src/utils/theme/theme.provider.tsx
```diff
@@ -0,0 +1,249 @@
...
+      tokens: themeTokens,
+      common: commonTokens,
+      setTheme,
+    }),
+    [themeState, setTheme],
+  );
+
+  return (
+    <ThemeContext.Provider value={themeContextValue}>
+      <ThemeBackgroundObserver displayTheme={displayTheme} />
```
pawanpaudel93: I think `ThemeBackgroundObserver` should be moved inside `StyledComponentsThemeProvider` as it uses `useTheme` hook and theme is `undefined` unless it's inside `StyledComponentsThemeProvider`'s `ThemeProvider` context and BE is breaking right now.

src/components/common/Image/Image.tsx
```diff
@@ -60,12 +61,21 @@ export function Image({
     [styles.isLoaded]: isLoaded,
   });
 
+  const placeholderURL =
```
pawanpaudel93, line 65: Shouldn't this be based on `displayTheme` not `themeMode` as if I set theme as `system` then themeMode is always `system` but my `displayTheme` might change based on system theme ?

src/utils/theme/theme.provider.tsx
```diff
@@ -0,0 +1,249 @@
...
+    function handleDarkModePreferenceChange(e: MediaQueryListEvent) {
+      setThemeState(getThemeState());
+    }
+
+    darkModePreference.addEventListener("change", handleDarkModePreferenceChange);
+
+    return () => darkModePreference.removeEventListener("change", handleDarkModePreferenceChange);
+  }, [themeMode, getThemeState]);
+
+  // (ONLY CONNECT, NOT BE) Add `data-theme` attribute based on resolved theme (light or dark, not system):
```
pawanpaudel93, line 153: Comment doesn't match the useEffect below.

src/utils/theme/theme.provider.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  useEffect(() => {
+    if (import.meta.env?.VITE_IS_EMBEDDED_APP !== "1") return;
+
+    document.documentElement.setAttribute("data-theme", displayTheme);
+  }, [displayTheme]);
+
+  // (ONLY CONNECT, NOT BE) Emit "THEME_CHANGE" event:
```
pawanpaudel93, line 162: Comment doesn't match the useEffect below.

src/components/common/Image/Image.tsx
Danziger: True, just fixed it. The other usages on this file use `themeMode` because `<source media="...">` takes care of that faster, but this one was a mistake.

src/utils/theme/theme.provider.tsx
Danziger: What do you mean? `import.meta.env?.VITE_IS_EMBEDDED_APP` is set to `"1"` only for Connect, so in BE both `useEffect`s will simply return here: `if (import.meta.env?.VITE_IS_EMBEDDED_APP !== "1") return;`

pawanpaudel93: Sorry, these two comments are incorrect and yes it runs on connect only.

pawanpaudel93 (approved):
tACK 🚀

PR #991, opened 2025-08-18
fix: arns requests

pawanpaudel93:
## Summary

This PR fixes the ArNS caching so there is no ArNS name requests frequently. Closes [ARC-1533](https://communitylabs.atlassian.net/browse/ARC-1533)

[ARC-1533]: https://communitylabs.atlassian.net/browse/ARC-1533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
+++ b/src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
@@ -8,7 +8,7 @@ import { getWallets } from "~wallets";
  */
 export async function handleSyncLabelsAlarm(alarm?: Alarms.Alarm) {
   // check alarm name if called from an alarm
-  if (!alarm || alarm.name !== "sync_labels") return;
+  if (alarm?.name !== "sync_labels") return;
 
   // get wallets
   const wallets = await getWallets();
--- a/src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
+++ b/src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
@@ -8,9 +8,7 @@ import { getWallets } from "~wallets";
  */
 export async function handleSyncLabelsAlarm(alarm?: Alarms.Alarm) {
   // check alarm name if called from an alarm
-  if (alarm && alarm.name !== "sync_labels") {
-    return;
-  }
+  if (!alarm || alarm.name !== "sync_labels") return;
 
   // get wallets
   const wallets = await getWallets();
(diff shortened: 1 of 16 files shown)


src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
```diff
@@ -8,17 +8,18 @@ import { getWallets } from "~wallets";
  */
 export async function handleSyncLabelsAlarm(alarm?: Alarms.Alarm) {
```
pawanpaudel93, line 9: This name cache syncing occurs every 1 hour for all the wallet the user has. Should we increase the interval at which we update the cache ?

src/lib/nameservice.ts
```diff
@@ -3,42 +3,39 @@ import type { NameServiceProfile } from "./types";
...
-    console.error(e);
+async function getProfileFromCache(walletAddress: string): Promise<NameServiceProfile | null> {
+  const cache = await initializeNameServiceCache();
+  const cachedProfile = cache[walletAddress];
+  return cachedProfile;
+}
+
+async function setProfileInCache(walletAddress: string, profile: NameServiceProfile | undefined) {
+  const cache = await initializeNameServiceCache();
+  cache[walletAddress] = profile || null;
```
pawanpaudel93, line 32: we store `null` now if we don't find ArNS name for the address. `none` was being stored before but there can be ArNS name as `none` as well.

src/api/background/handlers/browser/install/install.handler.ts
```diff
@@ -30,6 +30,9 @@ export async function handleInstall(details: Runtime.OnInstalledDetailsType) {
 
     // remove astro beta access announcement storage key
     ExtensionStorage.remove("astro_beta_access_announcement_shown");
+
```
pawanpaudel93, line 33: Clearing the cache as we store `null` instead of `none` if we don't find any name for the wallet address.

src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
```diff
@@ -8,17 +8,18 @@ import { getWallets } from "~wallets";
...
   // check alarm name if called from an alarm
-  if (alarm && alarm.name !== "sync_labels") {
-    return;
-  }
+  if (alarm?.name !== "sync_labels") return;
```
pawanpaudel93, line 11: I updated all the alarms check as this check was being skipped when sometimes `alarm` was `undefined`.

src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
7i7o: Yes, 1 hour seems too much. Let's start with 6 hours for now

7i7o (approved):
tACK 👍

src/api/background/handlers/alarms/sync-labels/sync-labels-alarm.handler.ts
pawanpaudel93: @7i7o Updated to 6 hrs https://github.com/wanderwallet/Wander/pull/991/commits/03f3025b9c78ccd649f157ab8fa5319de04a10ab

PR #993, opened 2025-08-18
fix: prevent token activity logo from shrinking

pawanpaudel93:
## Summary

This PR prevents some token activity logo from shrinking.

| Before | After |
|--------|-------|
| image | image |


7i7o (approved):
utACK 👍

PR #996, opened 2025-08-18
fix: Remove circular dependency introduced in the last release.

Danziger:
Doesn't seem to affect the production deployment, but Wander Connect doesn't run locally.


pawanpaudel93 (approved):
.

pawanpaudel93 (approved):
tACK 🚀

PR #1017, opened 2025-09-02
fix: swap quote

pawanpaudel93:
## Summary

This PR works on the following:
- Fix the issue where rapidly changing the quote input could result in the output not reflecting the most recent input.
- Update the network fee calculation based on wander fee
- Update calculating max value input


src/routes/popup/swap/utils/swap.hooks.ts
```diff
@@ -375,9 +411,9 @@ export function usePoolQuote({
     } finally {
       setIsLoading(false);
     }
-  }, [tokenIn, tokenOut, slippage, amountIn, poolId, poolType, wanderFeePercent]);
+  }, [tokenIn, tokenOut, slippage, amountIn, poolId, poolType, wanderFee, stopFetching]);
 
-  useAsyncEffect(async () => {
+  useEffect(() => {
```
pawanpaudel93, line 415: `useAsyncEffect` doesn't clean the interval on unmount so used the `useEffect` but we can fix the `useAsyncEffect` hook as well later for cleanup.

7i7o: Let's add a `TODO:` on the `useAsyncEffect` so it's documented

7i7o (approved):
tACK 👍

src/routes/popup/swap/utils/swap.hooks.ts
pawanpaudel93: Seems, there is already a TODO for that. We have some places that has cleanup function returned but it doesn't run on unmount.

pawanpaudel93: @7i7o Fixed the issue here: https://github.com/wanderwallet/Wander/pull/1018

src/routes/popup/swap/index.tsx
```diff
@@ -53,9 +54,14 @@ export function SwapView() {
...
   const [tokenSelectorType, setTokenSelectorType] = useState<TokenSelectorType>("send");
-  const { networkFee, isLoading: isNetworkFeeLoading } = useARNetworkFee({
+  const {
+    networkFee,
+    networkFeeValue,
+    isLoading: isNetworkFeeLoading,
+  } = useARNetworkFee({
     tokenIn: sendToken.processId,
     tokenOut: receiveToken.processId,
+    doubleFee: +defiFeeDetails?.finalFeePercent > 0,
```
pawanpaudel93, line 64: @7i7o We calculate double network fee here if there is wander fee so that we can cover the network fee when we send the wander fee.

src/routes/popup/swap/index.tsx
```diff
@@ -67,12 +73,17 @@ export function SwapView() {
     return toTokenBaseUnits(debouncedValueIn, sendToken.Denomination);
   }, [debouncedValueIn, sendToken]);
 
+  const maxValueIn = useMemo(
```
pawanpaudel93, line 76: @7i7o Max value calculation is updated as well so please check it as well by directly inputing the value or just hitting the max button.

src/routes/popup/swap/utils/swap.utils.ts
```diff
@@ -258,9 +258,9 @@ export function getSwapTime(poolType: PoolType | Provider) {
 }
 
 export function getPriceImpactColor(priceImpact: string, theme: DefaultTheme) {
```
pawanpaudel93, line 278: Updated this price impact color calculation as it was wrong and we need to show the red color when impact is greater than -10%

src/routes/popup/swap/index.tsx
7i7o: tACK 👍

src/routes/popup/swap/index.tsx
7i7o: tACK 👍

src/routes/popup/swap/utils/swap.utils.ts
7i7o: tACK 👍 
(you can see it with vAR > Game pool)

PR #1018, opened 2025-09-03
fix: useAsyncEffect cleanup on unmount

pawanpaudel93:
## Summary

Ensure async effects properly run cleanup on unmount and immediately if the async resolves after unmount, preventing memory leaks.


7i7o (approved):
ACK 👍

PR #1020, opened 2025-09-05
fix: improve swap history & others

pawanpaudel93:
## Summary

This PR introduces the following improvements:

- **Swap History**: Improve swap history. Uses local saved swaps data as well.
- **Swap Error Handling**: Detect swap errors and shows either the failed or completed screen accordingly. Added check for Permaswap invalid amount out errors.
- **Slippage Fix**: Corrects slippage calculation and use `minAmountOut` (which accounts for slippage) during swap execution — previously not used.  
- **Fee Processing**: Adds a safeguard to ensure Wander swap fees are processed only once, avoiding duplicate fee processing.

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -4387,12 +4387,16 @@
     "description": "Network/Provider fee text"
   },
   "swap_error": {
-    "message": "Swap error. Please try again.",
+    "message": "Swap failed. Please try again.",
     "description": "Generic error message shown when a swap transaction fails" 
   },
   "swap_error_increase_slippage": {
     "message": "Swap failed due to a price change. Try again increasing slippage.",
     "description": "Error message shown when a swap transaction fails due to invalid amount out"
+  },
+  "swap_error_permaswap_order": {
+    "message": "Failed to create permaswap order. Please try again.",
+    "description": "Error shown when permaswap order creation fails"
   },
   "free": {
     "message": "Free!",
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -4391,7 +4391,7 @@
     "description": "Generic error message shown when a swap transaction fails" 
   },
   "swap_error_increase_slippage": {
-    "message": "Swap error. Increase slippage and try again.",
+    "message": "Swap failed due to a price change. Try again increasing slippage.",
     "description": "Error message shown when a swap transaction fails due to invalid amount out"
   },
   "free": {
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -4390,6 +4390,10 @@
     "message": "Swap error. Please try again.",
     "description": "Generic error message shown when a swap transaction fails" 
   },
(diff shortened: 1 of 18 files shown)


assets/_locales/en/messages.json
```diff
@@ -4390,6 +4390,10 @@
     "message": "Swap error. Please try again.",
     "description": "Generic error message shown when a swap transaction fails" 
   },
+  "swap_error_increase_slippage": {
```
pawanpaudel93, line 4393: Maybe improve error messaging ? It's used for permaswap which is thrown if the expected amountOut is greater than the actual amount Out from permaswap.

src/routes/popup/swap/utils/alarms/swap-monitor/swap-fee-processor.ts
```diff
@@ -62,6 +70,12 @@ export async function processWanderFee(swap: SwapData): Promise<boolean> {
       return true;
     }
 
+    const isAlreadyProcessed = await checkIfWanderFeeAlreadyProcessed(swap);
```
pawanpaudel93, line 73: Double fee sending should not happen. But added this check to prevent that if it happens.

src/routes/popup/swap/utils/dex/dex.botega.ts
```diff
@@ -72,7 +72,7 @@ export async function getExpectedOutput({
   const protocolFeeQuantity = getTagValue("Protocol-Fee-Quantity", tags) || "0";
   const tokenInFee = BigNumber(lpFeeQuantity).plus(protocolFeeQuantity).toFixed(0, BigNumber.ROUND_DOWN);
   const minAmountOut = BigNumber(amountOut)
-    .multipliedBy(BigNumber(1).minus(slippage || 0))
+    .multipliedBy(BigNumber(100).minus(slippage || 0))
```
pawanpaudel93, line 75: Seems I calculated the `minAmountOut` wrong but this is not being used in the current prod version so it's fixed here and finally used in swap execution instead of `amountOut`.

src/routes/popup/swap/utils/dex/dex.permaswap.ts
```diff
@@ -218,6 +219,17 @@ export async function executeSwap({
       ],
     });
 
+    try {
```
pawanpaudel93, line 222: After doing the request order this message result can contain error if the `minAmountOut` is greater than the actual `amountOut` from Permaswap so catching the error early. You can actually reproduce this error by hardcoding a value greater than amountOut (amount before slippage deduction) and trying to do the swap.

assets/_locales/en/messages.json
7i7o: How about
`"Swap failed due to a price change. Try again increasing slippage."`?

7i7o (approved):
tACK 👍

assets/_locales/en/messages.json
pawanpaudel93: Let me update that. This message is better.

PR #1024, opened 2025-09-09
chore: add blocked OTP domains toast

7i7o:
This PR adds a notice for the email domains that have been blocking our OTP codes

Screenshot 2025-09-09 at 8 02 29 PM

### DIFF
--- a/src/routes/embedded/auth/auth/auth.view.tsx
+++ b/src/routes/embedded/auth/auth/auth.view.tsx
@@ -16,7 +16,7 @@ import {
 import React, { useCallback, useRef, useState } from "react";
 import { getSupabaseClient } from "~utils/embedded/embedded.utils";
 import { useLocation, useSearchParams } from "~wallets/router/router.utils";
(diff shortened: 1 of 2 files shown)


pawanpaudel93 (approved):
tACK 🚀

src/routes/embedded/auth/auth/auth.view.tsx
```diff
@@ -151,6 +151,18 @@ export function AuthEmbeddedView() {
           return;
         }
 
+        if (isOTPBlocked(email)) {
+          const blockedDomains = ["@icloud.com", "@me.com", "@mac.com"].join(", ");
```
pawanpaudel93, line 155:
```suggestion
          const blockedDomains = ["@icloud.com", "@me.com", "and @mac.com"].join(", ");
```

7i7o: Thanks for the suggestion
I left it as is on purpose since that's the copy Matt agreed on

PR #1029, opened 2025-09-11
fix: update collectibles query

pawanpaudel93:
## Summary

This PR updates the collectibles query to check if it's a valid collectible as the collectible verification is not working now.

### DIFF
--- a/src/tokens/aoTokens/sync.ts
+++ b/src/tokens/aoTokens/sync.ts
@@ -107,8 +107,7 @@ function getCollectiblesQuery() {
       tags: [
         { name: "Data-Protocol", values: ["ao"] },
         { name: "Type", values: ["Process"] },
-        { name: "Implements", values: ["ANS-110"] },
-        { name: "Content-Type" }
+        { name: "Implements", values: ["ANS-110"] }
       ]
       first: 100
     ) {


src/tokens/aoTokens/sync.ts
```diff
@@ -107,8 +107,7 @@ function getCollectiblesQuery() {
       tags: [
         { name: "Data-Protocol", values: ["ao"] },
         { name: "Type", values: ["Process"] },
-        { name: "Implements", values: ["ANS-110"] },
-        { name: "Content-Type" }
```
pawanpaudel93, line 111: We were checking `{ name: "Content-Type" }` tag presence as well with goldsky graphql but it is no longer working. You can check this query with and without it:

```query {
    transactions(
      ids: ["EcGIkIfGxSlucpKIzFj6JENE_GrNSQIEmETmE4gakZs"]
      tags: [
        { name: "Data-Protocol", values: ["ao"] },
        { name: "Type", values: ["Process"] },
        { name: "Implements", values: ["ANS-110"] },
        # { name: "Content-Type" }
      ]
      first: 100
    ) {
      pageInfo {
        hasNextPage
      }
      edges {
        node {
          id
          tags {
            name,
            value
          }
        }
      }
    }
  }```

7i7o (approved):
tACK 👍

PR #1034, opened 2025-09-15
fix: frequent reload or black screen due to unpartitioned status

pawanpaudel93:
## Summary  
This PR resolves the issue where Wander Connect repeatedly reloads when **unpartitioned storage access** is not granted.  

## How to test  
1. In Brave, ensure unpartitioned storage access is not granted (or revoke it if already granted).  
2. Open the **prod URL** and the **PR URL** separately.  
3. Reload the tab multiple times and observe the behavior:  
   - **Before**: Wander Connect UI reloads repeatedly.  
   - **After (with this PR)**: Reloading issue is fixed.  

## TODO  
- [X] Remove development logs before merging.


7i7o (approved):
tACK 👍

PR #1038, opened 2025-09-18
refactor: add non-transferable token whitelist check

pawanpaudel93:
## Summary

This PR adds the whitelist check for the non-transferable tokens.


7i7o (approved):
tACK 👍

PR #1047, opened 2025-10-06
fix: use default CU for wAR token balance

pawanpaudel93:
## Summary

This PR updates wAR to use the default CU instead of [https://ao.arweave.asia](https://ao.arweave.asia), as the token balance loads properly with default CU. Also, Permaswap and AOX also no longer seem to use [https://ao.arweave.asia](https://ao.arweave.asia).


7i7o (approved):
tACK 👍

PR #1058, opened 2025-10-27
fix: firefox build

pawanpaudel93:
## Summary

This PR fixes an issue in the Firefox build where the auth request popup could not open due to `@arconnect/webext-bridge` not getting the correct context. It also includes additional minor fixes.

**Related PR:** [wanderwallet/webext-bridge#1](https://github.com/wanderwallet/webext-bridge/pull/1)

---

## How to Test

1. Since Firefox MV3 config is used here, the **development build** currently does not function correctly. I think it might be due to firefox being experimental for MV3 https://docs.plasmo.com/framework/workflows/faq#what-are-the-officially-supported-browser-targets  
2. The **production build**, however, works as expected.  
   - Build the production version.  
   - Load it manually in Firefox and verify that the auth request popup opens properly & others work as expected as well.  
3. (Optional) We can switch to **Manifest V2** and its working properly in both development and production builds.


package.json
```diff
@@ -62,7 +62,7 @@
     "@arconnect/components": "^1.0.0",
     "@arconnect/components-rebrand": "https://github.com/wanderwallet/components-rebrand.git#master",
     "@arconnect/keystone-sdk": "^0.0.5",
-    "@arconnect/webext-bridge": "^5.0.6",
+    "@arconnect/webext-bridge": "https://arweave.net/37OdN5bUTjIFRScBJn0ViGyJn0F935GbMgoe-tqxa1E",
```
pawanpaudel93: Update `@arconnect/webext-bridge` package and use the updated package here.

7i7o (approved):
tACK 👍 
🦸

package.json
pawanpaudel93: Updated `@wanderapp/webext-bridge` package in commit: https://github.com/wanderwallet/Wander/pull/1058/commits/6fd4aa32a74bb6d1c57545546dd89f9d0f68f3ab

7i7o (approved):
tACK 👍 

Re-tested with newly deployed webext package

PR #1059, opened 2025-10-27
feat: add cloud backup to connect

pawanpaudel93:
## Summary

This PR adds the feature to backup wallet to cloud (google drive & iCloud) & import it. It also has updated sign-in with OTP screen.


7i7o (approved):
tACK 👍

Now we only need to update the credentials for each cloud service

PR #1060, opened 2025-10-30
fix: sdk build

pawanpaudel93:
## Summary

This PR updates the tsup config to reduce the build size while building.

### Build size improvement for npm
The changes made in tsup config:
- Minify code for production with changes made in `package.json` scripts.
- Use single file `src/index.ts` in entry (Didn't see usage of multiple entries `src/**/*.ts` as modular import doesn't seem to be needed)
- Separate configuration for `iife` so we can disable sourcemap & dts.

#### Build sizes
- New build size: `1.7MB`
- Previous build size: `6MB`


7i7o (approved):
tACK 👍

PR #1075, opened 2025-11-12
refactor: remove ardrive cu from balance dryrun

pawanpaudel93:
## Summary

This PR removes WNDR & USDA process from using ardrive cu and leaves ARIO token processes as it is.

### DIFF
--- a/src/tokens/aoTokens/ao.ts
+++ b/src/tokens/aoTokens/ao.ts
@@ -16,17 +16,8 @@ import BigNumber from "bignumber.js";
 import { CACHE_API } from "~constants/api";
 import Arweave from "arweave";
 import { queryClient } from "~utils/tanstack";
-import {
-  USDA_PROCESS_ID,
-  Id,
-  Owner,
-  AR_PROCESS_ID,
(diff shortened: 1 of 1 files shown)


src/tokens/aoTokens/ao.ts
```diff
@@ -61,7 +61,7 @@ export const DEFAULT_CU_URL = "https://cu.ao-testnet.xyz";
 const { dryrun: arDriveDryrun } = connect({ CU_URL: ARDRIVE_CU_URL });
 const { dryrun: aoDevDryrun } = connect({ CU_URL: AO_DEV_CU_URL });
 
-const ARDRIVE_PROCESSES = [ARIO_MAINNET_PROCESS_ID, ARIO_TESTNET_PROCESS_ID, WNDR_PROCESS_ID, USDA_PROCESS_ID];
```
pawanpaudel93, line 64: should we stop using cu ardrive even for ARIO processes ?

7i7o: Yes

7i7o (approved):
tACK 👍

PR #1081, opened 2025-11-26
fix: add custom cu for ARIO

pawanpaudel93:
## Summary

This PR restores the use of the custom CU for ARIO token.


7i7o (approved):
tACK 👍

PR #1084, opened 2025-12-02
fix: tx history

pawanpaudel93:
## Summary

This PR stores pending transactions and displays them from storage until they are received from GraphQL, at which point they are cleaned up if they are stored more than 5 minutes. It also adds cleanup for pending transactions that are 1 day or older and still hasn't been available on graphql.

## How to Test
1. Send token from one wallet to another. It can be wallets within Wander also.
2. Check that token activity by clicking on the token & Feed as well for seeing the pending transactions.


src/tokens/hooks/index.ts
```diff
@@ -88,8 +89,8 @@ export function useTokenBalance(token: TokenInfo | null, address: string, refres
     queryFn: async () => {
       try {
         const balance = await retryWithDelay(
-          () => fetchTokenBalance(token, address, refresh),
-          4,
+          () => timeoutPromise(fetchTokenBalance(token, address, refresh), 10000),
```
pawanpaudel93, line 92: Added 10s timeout and reduced balance fetch count from 4 to 3.

7i7o (approved):
tACK 👍

PR #1087, opened 2025-12-05
fix: token balance refetch

pawanpaudel93:
## Summary

This fixes the issue where the token balance showed `0` after sending a token. The problem was caused by how the token balance query was being invalidated.

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -108,7 +108,16 @@ export function formatBalance(balance: BigNumber | string) {
   const tooltipBalance = balance.toFormat(20, BigNumber.ROUND_FLOOR).replace(/\.?0*$/, "");
   let suffix = "";
 
-  if (balance.lte(1e4)) {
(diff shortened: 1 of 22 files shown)


src/utils/format.ts
```diff
@@ -108,7 +108,16 @@ export function formatBalance(balance: BigNumber | string) {
   const tooltipBalance = balance.toFormat(20, BigNumber.ROUND_FLOOR).replace(/\.?0*$/, "");
   let suffix = "";
 
-  if (balance.lte(1e4)) {
+  if (balance.isZero()) {
+    displayBalance = "0";
+    showTooltip = false;
+  } else if (balance.lt(1)) {
```
pawanpaudel93, line 114: Added this so that very small balances are shown 0 if decimals greater than 12.

7i7o (approved):
tACK 👍

PR #1107, opened 2025-12-18
refactor: migrate analytics to google

pawanpaudel93:
## Summary

This PR migrates analytics tracking from Segment to Google Analytics (GA4), following the official Chrome Extensions integration guide: [https://developer.chrome.com/docs/extensions/how-to/integrate/google-analytics-4](https://developer.chrome.com/docs/extensions/how-to/integrate/google-analytics-4)


7i7o (approved):
tACK 👍

src/utils/analytics.ts
```diff
@@ -131,69 +137,220 @@ export enum PageType {
...
 
     // first we check if we are allowed to collect data
     const enabled = await getSetting("analytics").getValue();
-
     if (!enabled) return;
 
     const ONE_HOUR_IN_MS = 3600000;
 
     // TODO:login is tracked only once and compared to an hour period before logged as another Login event
     if (eventName === EventType.LOGIN) {
```
pawanpaudel93, line 385: @7i7o should we use local or session storage for login event as currently we use session and the stored data is lost when browser is closed ? wdyt ?

